Issue #78, opened 2016-12-22
Demonstrate multiple stack layers


neilhunt1:
Create two stacks, one with simple infrastructure such as VPC's, subnets, security groups; another with application-specific resources such as EC2 instances, S3 buckets, RDS, ElastiCache, etc.
This should demonstrate the ability to separate different resources within an application's spectrum and can support "separation of duties" (e.g. Sec owns security groups, dev team owns it's own databases, ops team owns common S3 buckets, etc.).

robproper:
In progress, but need to share output from infrastructure to pilot stack (automatically)

PR #87, opened 2017-01-06
Terraform pilot

robproper:
These changes allow one or more stack input files to be specified to the manage-stack.sh command. The referenced input file must be a valid [Terraform variable file](https://www.terraform.io/docs/configuration/variables.html#variable-files), so either an HCL (tfvars) or json formatted file. 

The input argument may specify:

1. An S3 URI, e.g. `--input s3://my-bucket/path/to/input.tfvars`. This assume that the user running the script has access rights to this S3 object (either the object is publicly accessible, or the used AW-CLI profile has access).
2. A HTTP/S URL e.g. `--input https://s3.amazonaws.com/my-bucket/path/to/input.tfvar`. This assume that the user running the script has access rights to this S3 object (either the object is publicly accessible, or the used AW-CLI profile has access)
3. A stack output URI, e.g. `--input stack-output://system-name/stack-name/branch-name`.  This allows propagation of variables from one to another stack. The output generated (with manage-stack.sh) for another stack will be downloaded.  This allows propagating  output from one stack to the next . Note that system-name can be empty, in which case the default "pilot" will be used". It assume the same bucket (as configured for the using stack) holds the generated output file. Generated output files will always be HCL formatted (tfvars) files.

Further note that:

- The current script only handles generation of string variables, not structured (array and map) variables. The Terraform [output cli command](https://www.terraform.io/docs/commands/output.html) does not generate valid HCL, so the script converts the output. This should be added at a later time.
- Ideally a dependency between stack could be defined, such that the manage-stack script can manage the relationship between stacks. This could be added as a command-line argument (e.g. --depends-on stack-name/branch-name). This should be added at a later time.

Issue: #78


neilhunt1 (commented):
I like it. So this essentially allows us to automatically send our terraform output from one stack to a location and automatically ingest it in another stack provided we reference the same s3 bucket as the output param of the first as the input param of the second? Essentially it allows us to interchange tfvar files between stacks in an S3 bucket?

robproper:
Correct. Or, if there is part of the infrastructure  (e.g. the VPC) that has been created in a different way, we can store its information in a tfvars file in an S3 bucket as well

neilhunt1:
"create in a different way" you mean some non-terraform way such as console, CloudFormation, cli, etc. that would not automatically output a tfvars file, we would simply manually translate any needed info to a tfvars file ourselves and store it in the same way in an s3 bucket and our own terraform scripts would act the same?

robproper:
Correct, so we can support both cases.

Issue #82, opened 2016-12-22
Trigger network infrastructure pipeline from Jenkins


neilhunt1:
Address issues and demonstrate ability to setup infrastructure using Jenkins Pipeline and understand how it relates to Resources and Ansible pipelines.

robproper:
Use outcome of #78 and #79 to have orchestrate creation of different stacks called from Jenkins. 

Stacks can be created in in a single pipeline or multiple pipelines. Multiple pipelines (likely) requires separate repositories for separate stacks (or else use a parameterized pipeline to enable naming the stack to create in the pipeline job (if using a single repository).

PR #88, opened 2017-01-06
82 multistack pipeline

robproper:
Changed (main) Jenkins pipeline to:

1.Create an infrastructure stack, using the branch name
2 Create a pilot stack (bastion and monitor instances) using subnet and security group output from the infrastructure stack, also using on the (same) branch name.

NOTE: To get Jenkins to build infrastructure for a specific branch (using the Provisioning Pipeline) add the branch name in the included branches. This is to avoid building infrastructure for all unrelated branches, as this repository has a multi-purpose.

#82

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,11 +17,14 @@ node("master") {
 def manageStack(stack_name, branch_name, dependsOn = null) {
     dir("terraform/") {
         def script = "${pwd()}/bin/manage-stack.sh"
-        def args = "-v -v --action create"
+        def args = []
         sh "chmod 700 '${script}'"
+        args << "-v"
+        args << "-v"
+        args << "--action create"
         if (dependsOn) {
-            args += "--input stack-output:///infrastructure/${env.BRANCH_NAME}"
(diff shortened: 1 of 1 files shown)


neilhunt1:
Question, what happens if we want to separate these two stacks (inf and pilot)? Say two teams want to use the same VPC (inf stack)?

robproper:
Easiest would be to create dedicated repositories, because this allows for a dedicated Jenkinsfile and dedicated branches. That is, feature branches are all about the one stack, whereas with multiple stacks in one repository branches could be about unrelated features. That requires Jenkins multi-branch pipelines to be updated to explicitly specify the applicable branches (which in undesirable).

If we keep one repository, we would have to customize the Jenkinsfile to trigger a specific set of tasks specific to the stack (like we discussed before), e.g. by including another dedicated Jenkinsfile based on the Jenkins job. Additionally (or even alternatively) we could have a dedicate configuration file in a known S3 bucket to further configure the dedicate pipeline for the stack.

neilhunt1:
Makes sense to me. Yes we will likely continue to find needs to isolate git repositories by pipeline and this approach should solve that if we decide we want to go that way.

PR #93, opened 2017-01-12
79 terraform ansible

robproper:
Added the following changes:

1. The ability to invoke Ansible playbooks from the pipeline

2. The ability to specify for a terraform "stack" to wait for resources to complete
    Create a file name stack.yml in root module source directory of the stack, using the following template:

```
stack:
       wait:
             conditions:
                  <resource name>:   aws-healthcheck
             delay: 10
             max-iterations: 60
```

where:

- `conditions` is a map of all resources to wait for
- `<resource name>`  is the value of the Resource tag created in the Terraform script.
- `aws-healthcheck` indicates to wait for the ec2-instance to pass both health checks
(other checks can be added later, e.g. to wait for response to HTTP requests on a specific port could be represented as `http 80 /path`)
- `delay` is the waiting period in seconds in between checks. The default is `10`.
- `max-iterations` is the maximum number of iteration over all resources. The default is `720`, which equals 2 hours.

4. Added `stack.yml` files for both the infrastructure stack (to wait for NAT to pass health checks) and 
   the pilot stack (to wait for the bastion and wordpress-web instances to pass health checks).

3. Added Ansible playbook invocations of jumpbox.yml and datagov-web.yml playbooks against the 
    bastion resp. wordpress-web resources created by the pilot stack. Not using dynamic inventory yet.


terraform/bin/manage-stack.sh
```diff
@@ -144,6 +151,107 @@ function s3_bucket_exists {
...
+    local prefix=$2
+    if  [ ! -f "${file_name}" ]; then
+       echo "File '${file_name}' does not exist" >&2
+       exit 1
+    fi
+    debug "config-data(file):\n$(cat $file_name)\n"
+    cat $file_name | parse_yaml_from_string
+}
+
+function parse_yaml_from_string {
```
neilhunt1, line 192: Assume you borrowed this from [here](http://stackoverflow.com/questions/5014632/how-can-i-parse-a-yaml-file-from-a-linux-shell-script)? Looks good.

terraform/bin/manage-stack.sh
```diff
@@ -493,14 +619,105 @@ function do_stack {
...
+                     "Name=tag:Resource,Values=${resource}" \
+                     "Name=instance-state-name,Values=running" \
+            --query "Reservations[].Instances[].{Id:InstanceId}" \
+            --output text)
+    if [ "${instanceID}" == "" ]; then
+        debug "Resource ${resource} not found"
+        return 1
+    fi
+    # Check the health-check status
+    status=$(ec2 describe-instance-status \
```
neilhunt1, line 647: Does this strictly check for SSH being available? Do you know what else, if anything, it may be checking?

terraform/bin/manage-stack.sh
```diff
@@ -493,14 +619,105 @@ function do_stack {
...
+        --output text)
+    if [ "${status}" == "" ]; then
+        debug "Resource ${resource} health checks not passed"
+        return 1
+    fi
+    debug "Resource ${resource} health checks passed"
+    return 0
+}
+
+function all_resources_completed {
```
neilhunt1, line 662: so effectively this allows the terraform script to check that all resources are completed including any health checks? This is what will be used to tell ansible to proceed in its execution?

terraform/pilot/main.tf
```diff
@@ -15,15 +15,20 @@ variable "branch" {}
...
+    type = "map"
+    default = {
+        region = "us-east-1" 
+    }
+}
+variable "ami_type" { default = "not-hardened" }
 variable "bastion_subnet_id" {}
 variable "bastion_security_group_id" {}
-variable "bastion_instance_type" {}
+variable "bastion_instance_type" { default = "m3.medium" }
```
neilhunt1, line 27: are they using m3.medium now? Or how did we determine this size? Just curious.

neilhunt1 (approved):

JJediny:
@aquilent-robproper @neilhunt1 LGTM merge when ready

terraform/bin/manage-stack.sh
robproper: yes, "borrowed" from that source, and before that borrowed from other code I had already written

terraform/bin/manage-stack.sh
robproper: "Strictly" speaking it checks that the AWS health checks pass. You cannot use the AWS-CLI to perform OS level availability. Typically, this means that the SSH is available, but if necessary we can add to ability to perform an additional check that attempts to connect using SSH.

terraform/bin/manage-stack.sh
robproper: At this point it allows one to specify which named resources need to pass AWS health checks (as this is currently the only wait condition supported) before proceeding.

terraform/pilot/main.tf
robproper: Not sure. Typically, AWS requires at least medium instance types to allow you to perform security scans, so this is a "sensible" default, but perhaps another instances type makes more sense.

JJediny: generally speaking for our compute heavy applications c4.xlarge or c4.2xlarge for mem heavy m4.xlarge/m4.2xlarge for prod loads for demo/dev **c4.large and m4.large**

neilhunt1: Yep that's the formula Dan and I have been using for many of the app servers. But what if it's not anything-heavy as would likely be the case for bastion server? t2.medium?

neilhunt1: According to [this](https://docs.google.com/spreadsheets/d/1YLiMhKgIMdqvTxlvjwA5MKfqlY-53oOyXm1VMrOoruE/edit#gid=200356483), looks like for our smallest instances we are using t2.small for dev, t2.medium for prod.

PR #101, opened 2017-01-18
94 terraform environments (update)

robproper:
Small changes to improve the pipeline in the master (master-demo) branch

#96

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,19 +18,19 @@ if (isMaster()) {
 }
 
 def provision(environment) {
-    stage(environment ': Provision Infrastructure') {
+    stage("${environment}: Provision Infrastructure") {
         node("master") {
             runTerraform('infrastructure', environment)
         }
     }
 
-    stage(environment ': Provision Pilot') {
+    stage("${environment}: Provision Pilo") {
         node("master") {
             runTerraform('pilot', environment, "infrastructure")
         }
     }
 
-    stage(environment ': Provision Jumpbox') {
+    stage("${environment}: Provision Jumpbox") {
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -9,32 +9,42 @@ stage('Initialize') {
...
 }
 
 def provision(environment) {
-    stage('Provision Infrastructure Stack') {
+    def label = ((environment.startsWith("dev")) ? "dev" : environment).
+        toUpperCase();
+
+    if (label != "DEV") {
+        stage("master") {
+            timeout(time:5, unit:'DAYS') {
```
neilhunt1, line 21: Just curious the "timeout" command - I think this is a newer one they added to the syntax? So it kills the pipeline run if no action is taken by the end of the timeout? And you chose 5 days as sort of an arbitrary number?

Jenkinsfile
```diff
@@ -9,32 +9,42 @@ stage('Initialize') {
...
+                input "Deploy to ${label}?"
```
neilhunt1, line 22: yep, input should be outside a node, otherwise the node will remain running indefinitely

neilhunt1 (approved):
Looks good

Jenkinsfile
robproper: Yep, 5 days is arbitrary, although I simply copied it from [Top 10 Best Practices for Jenkins Pipeline Plugin](https://www.cloudbees.com/blog/top-10-best-practices-jenkins-pipeline-plugin). Likely a smaller number makes more sense, but it can always be adjusted later.

PR #105, opened 2017-01-20
96 manage cf

robproper:
Implemented the ability to manage provisioning using cloud-formation:

1. Create a manage-stack.sh script, similar to the Terraform management script. Takes a stack and environment name as argument.
2. Use S3 to push cloud-formation and other supporting scripts
3. Allow the use of Terraform (tfvars) input files to specify stack parameters, that can be specified as a local file name, an S3 URI, or a HTTP(S)-based URL.
4. Script determine whether to create or update and waits for stack action (create/update) to complete
5. Can specify in a YAML-based configuration file the following:
     - A list of stack parameters, with values
     - Stack parameter values may have environment variable placeholder. The script substitutes the placeholders with the value of the environment variable.
     - A mapping of map input file property names to stack parameter names (names will not always match). 
    - Which input properties to ignore when generating stack parameters.
      NOTE: Should probably change this to explicitly specify all parameters, where a parameter
      can be specified as the value of a input property (similar to b)
    - List a number of scripts to be uploaded into S3, that can subsequently be used in the cf-init stage of AWS resources
    - A list of EC2 instances (by logical stack id) to wait for after the stack completes. The script will wait until the instance's EC2 health checks pass

Change the Jenkins pipeline to:

1.  Be able to create the full pipeline, or specific (parts of the) pipeline (datagov-terraform, datagov-ansible, d2d bastion). The various 'pipeline' are in their own groovy file (under jenkins/pipeline) that once loaded will return a 'pipeline' object that can has implementations of various stages: initialize, provision, test, ....
2. Allow the main Jenkins file to discover which pipeline to run based on the JOB_NAME
3. If using the full pipeline it calls the other scripts
4. Separate invocation of the terraform and cloud-formation management script, as well as Ansible into their own groovy file (under jenkins/provisioner)

NOTES:
1. The main Jenkins file defines the stages, but will call the pipeline designated by phrases in the JOB_NAME, so essentially all pipelines follow a similar pattern: initialize, provision, test, ... others to be determined later. The loading of another pipeline only works inside a Stage/Node (so it was necessary)
Ideally the individual pipelines are change to objects using the same base class that can implement the default implementation of stages ('do nothing'), such that if new stages need to be added they do not immediately have to be changed in all pipelines.

2. Under "normal" circumstances pipelines should be defined in separate repositories, such that the pipeline can run automatically upon commits/merges. If sharing a repo that would be highly inefficient, because changes that are checked in may not have any bearing on the designated pipeline.

3. The 'delete' action does NOT remove the S3 bucket content for the stack/environment . Should be added later.


cloud-formation/infrastructure/stack.yml
```diff
@@ -0,0 +1,19 @@
+stack:
```
neilhunt1, line 1: Not sure about these ^M signs? Win/Unix line ending issue?

cloud-formation/d2dbastion/stack.yml
```diff
@@ -0,0 +1,21 @@
+stack:
```
neilhunt1, line 1: Is this your own proprietary yml syntax? Will people have to learn this to use these scripts?

cloud-formation/d2dbastion/scripts/bastion_bootstrap.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
```
neilhunt1, line 1: I'm unclear what this is doing specific to a bastion server? Or is this something that should be applied to any CloudFormation-provisioned EC2 instances?

neilhunt1 (approved):
Well done

cloud-formation/infrastructure/stack.yml
robproper: I will have to update that file

PR #107, opened 2017-01-23
fix typos in terraform doc

afeld:


robproper:
LGTM

afeld:
I set the target as `master-demo`...was that right? Will it get into `master`?

Issue #109, opened 2017-01-24
Make down-stream feature branche pipelines use "shared" dev environment


robproper:
Currently, when creating a feature branch in a down-stream pipeline it assumes the upstream environment exists for hat particular branch (i.e. an environment named `dev-<branch-name>`). This should be change to make it always use the downstream `dev` environment instead

PR #110, opened 2017-01-24
109 downstream dev

robproper:
Changed upstream pipeline reference to use dev environment in both the cloud-formation and the terraform groovy scripts when using a feature branch (i.e. environment is dev-<branch-name>)

#109


robproper:
Note this includes the dynamic inventory changes, so should review pull request #108 first

PR #115, opened 2017-01-25
114 change playbook

robproper:
Changed the wordpress-web playbook invocation to wordpress-web-demo which then only installs nginx


neilhunt1 (commented):
LGTM

PR #136, opened 2017-03-23
Non existing groups

vasili4:
Build fails when `inventory_hostname in groups["kibana"]` is used in condition, and group named `kibana` does not exist.
I've both added empty groups to staging inventory, and added additional checks like `if "kibana" in groups` 
In future any of those steps is required to avoid errors. I believe first one (with empty groups in inventories) is preferred.


dano-reisys (commented):
thx ;)

dano-reisys (approved):

PR #146, opened 2017-04-08
log real ip for ckan apache

FuhuXia:


vasili4 (approved):
must work. Official recommendation is to use mod_remoteip though

PR #170, opened 2017-05-24
add cron to clean solr catalina log files before filling up disk

FuhuXia:


vasili4 (approved):

dano-reisys:
o_O

JJediny:
@FuhuXia confused are we using cron or logrotate for this, looks like cron but other log handling is done with logrotate if others are ok LGTM just raising this as an item

FuhuXia:
Tomcat has its own mechanism to rotate files. Using logrotate on top of is awkward. I'd rather use a simple cron for it.

https://stackoverflow.com/questions/8962477/logrotate-files-with-date-in-the-file-name

dano-reisys:
I'd rather use one consistent mechanism for all apps + control the behavior of each log through one playbook.

FuhuXia:
The bad news is that if we use logrotate with default tomcats config, the partition will be filled up 100% by those catalina.yyyymmdd.log files. We have two options, 1) either add a new tomcat6 config file to the playbook to change default tomcat behavior, or 2) use a shell script as cronjob or logrotate/postrotate script to delete those tomcat6 daily log files.

This commit is using cronjob approach. I will modify it to use a tomcat config file approach. Will commit it to the same pull request.

PR #186, opened 2017-06-15
ansible cfg fix

vasili4:


JJediny:
@alex-perfilov-reisys wish roles_path allowed for multi-line values but otherwise LGTM

vasili4:
@JJediny yes, updating long lines is hard to track. Just a bit easier than updating vaults :D I was also thinking of creating symlink ansible/ansible -> ansible, to avoid duplicating those paths :) But that may introduce even more mess. This is done to enable running playbooks from root folder of this repo, instead of going to /ansible folder ( I know at least one person who prefers that way ;) )

vasili4:
it was merged by #196

FuhuXia (approved):

PR #187, opened 2017-06-19
Dup headers ckan

vasili4:


ansible/roles/software/ckan/security-fix/tasks/main.yml
```diff
@@ -10,3 +10,21 @@
   with_items:
     - /etc/ckan
     - "{{ project_source_new_code_path }}"
+
+- name: Fix 'Content-Cache' header
+  replace:
+    dest: "{{ project_source_new_code_path }}/src/ckan/ckan/lib/base.py"
+    regexp: ' = "private"'
```
JJediny: is replace or lineinfile safer/more reliable?

JJediny (commented):
LGTM for history - what is our preference for which to use/when lineinfile vs. replace?

ansible/roles/software/ckan/security-fix/tasks/main.yml
vasili4: The benefit of lineinfile is that it will add required line even if it doesn't find the pattern. 
There is also [blockinfile](http://docs.ansible.com/ansible/blockinfile_module.html) module for multiple lines.

JJediny: @alex-perfilov-reisys Should we not be copying over the files completely/explicitly? https://github.com/GSA/d2d/pull/400

vasili4: @JJediny , looks like I do not have access to that repo / invalid url

JJediny:
```
Currently, we're having Ansible find and replace single lines in the existing sshd_config, which can get messy. The file I looked at on one of the jump boxes has a bunch of duplicate lines, which makes it difficult to follow, and can lead to bugs.

This pull request replaces all of that logic by overwriting the ssd_config from a template. This is a practice I saw recommended by various Ansible users, as it makes the resulting state of the file deterministic/explicit.

Would love early feedback!
```
LGTM as is (lets queue up when/what in another issue for template management) - @dano-reisys @FuhuXia please merge

vasili4: For major file updates that would be legit. I don't believe it's possible to introduce duplicates with `lineinfile` using `regexp: '^SomeSetting ' ` though

FuhuXia: if copy the whole fine, it is easy to lose track what and why the change we made over the defaults. when doing replace I would prefer including more text into the regex phrase. It is a file with a couple of hundred lines and it might contains multiple occurrences of ` = "private"` in the future version.

so i prefer lineinfile with more defined regex phrase.

JJediny: +1 thanks for elaborating

vasili4: that is possible. `lineinfile` updates only 1 last found regex match though, not necessary the correct one )

dano-reisys (approved):

PR #199, opened 2017-06-22
No apache2 on harvester

vasili4:


FuhuXia:
We have a playbook for ckan apache. 
https://github.com/GSA/catalog-deploy/tree/master/ansible/roles/software/ckan/apache
do you want to move the code there?

vasili4:
one of the steps changes code at `src/ckan/ckan/lib/base.py`. If I move it to apache role, the further code deployment will rewrite my change.

FuhuXia:
I see. any reason why the code change does not go to ckan repo but modified here with playbook?

https://github.com/GSA/ckan/tree/2.3-upgrade

vasili4:
to simplify future core updates of course

FuhuXia:
Ok. as long as we remember we modify the code in the playbook, and a few months down the line we don't scratch our head wondering why the ckan code on the server does not show up clean with 'git status' check.

FuhuXia (approved):

PR #202, opened 2017-06-27
Getting started on kitchen-docker for ansible

JJediny:


vasili4 (approved):
rebased to latest master

PR #208, opened 2017-07-18
Updates to README and Badge

JJediny:


vasili4 (approved):
let's try keep that badge green 👍

PR #241, opened 2017-09-25
grouo tags for bsp and ami-fix related

FuhuXia:


dano-reisys (commented):
I've fixed the ami permissions at the ami level, not sure if the ami-fix is needed anymore. if it is needed it should be applied to all apps not just catalog.

dano-reisys (approved):

PR #268, opened 2017-10-27
set_fact: worker_type changed from str to var

John123Yu:


FuhuXia (approved):
Thanks for catching it.

PR #275, opened 2017-11-02
Remove vulnerable swfupload.swf file 

anup-khanal:


ansible/roles/software/wordpress/datagov-sudo-2-init/templates/datagov.nginx.conf
```diff
@@ -68,7 +68,7 @@ server {
         log_not_found off;
         access_log off;
     }
-
+
```
vasili4, line 71: this can be removed from PR

PR #290, opened 2017-12-14
Redirect Rules for Data.gov

anup-khanal:
Moving away from Akamai redirects.


anup-khanal:
added 302 redirects as well

PR #363, opened 2018-05-18
Removing duplicative cron job on dashboard

philipashlock:
It looks like the `datajson` cron was colliding with the `full-scan` job so that the `full-scan` results weren't captured. However, since `full-scan` includes all the operations provided by `datajson` it should be the only one needed.


JJediny (approved):
LGTM

PR #375, opened 2018-08-07
Add playbook for one-off patch

adborden:
Includes the patch file and playbook for https://github.com/GSA/ckan/pull/19


philipashlock (approved):
Looks good. Thanks!

PR #376, opened 2018-08-07
updates to patch

philipashlock:


philipashlock:
This was done with #377 instead

PR #378, opened 2018-08-07
ckan.site_url should include protocol

adborden:
http://docs.ckan.org/en/2.8/maintaining/configuration.html#ckan-site-url

Without the protocol, the site_url doesn't seem to be used at all.


adborden:
@JJediny all the other URLs were correct, but in some cases we are pulling from vault. We should verify that those vault values also include the protocol.

@hareeshreddyg do you know where the value comes from for the dev environment? we manually had to fix that, but I'm not sure where that exists in the ansible roles.

JJediny (approved):
OK

PR #383, opened 2018-08-15
Add redis password to group vaults, vars, and application config

jasonschulte:
This was tested by running the catalog.yml playbook locally with tags that only install and configure redis.  

`ansible-playbook catalog.yml -i inventories/local/localhost --tags="harvester" --skip-tags="frontend"`

After running the playbook, password auth was verified with `redis-cli -a <PASSWORD> info`


JJediny (approved):
LGTM - no exposed secrets reviewed

PR #384, opened 2018-08-16
Enable kitchen for local development

adborden:
This adds kitchen to CircleCI and enables it for local development.


ansible/roles/software/catalog/ckan-app/templates/etc/ckan/production.ini.j2
```diff
@@ -204,7 +204,9 @@ ckan.tracking_enabled = True
 set debug = False
 ckan.harvest.mq.type = redis
 ckan.harvest.mq.hostname = localhost
+{% if redis_password is defined %}
```
adborden: Spoke to @JJediny about this and we should just set the variable in all the ckan based inventories

adborden:
@JJediny this is ready for review

JJediny (approved):
LGTM

PR #394, opened 2018-08-31
modify nginx config to redirect all HTTP traffic to HTTPS

jasonschulte:


JJediny (approved):
needs to be rebased but good

adborden:
This only needs to be applied to BSP dev. This should be re-applied to confirm and close.

PR #395, opened 2018-08-31
postgis upgrade playbook

adborden:


adborden:
Not really needed anymore

PR #396, opened 2018-09-04
Terraform sandbox for testing

adborden:
Very beginnings of terraform templates to create a sandbox to test. I used it to test the RDS upgrades. It doesn't setup much beyond that.


JJediny (approved):
We'll want to move this to https://github.com/gsa?utf8=%E2%9C%93&q=datagov-infrastructure&type=&language= but OK for testing

PR #401, opened 2018-09-10
Redis install dir

adborden:
Updates redis install directory to `/usr/local` to avoid permissions issue in `/opt` with the hardening script.


JJediny (approved):
Approved

PR #405, opened 2018-09-13
Site url fixes for catalog

adborden:
Dev/test instances are not setting the site_url correctly, they are defaulting to `https://catalog.data.gov`. Each environment has their own unique URL.

cc @hareeshreddyg


JJediny (approved):
👍

PR #406, opened 2018-09-17
[WIP] install gsa hardening role

JJediny:


adborden (approved):
:+1: 

It might be easier to test this if we split out updating the other vendored roles next time. If any of the roles introduce a regression, it might be difficult to track down with so many changes.

adborden:
@JJediny does this replace #379 ?

adborden:
@hareeshreddyg and I tested this out today and ran into a few issues.
- The hardening playbook updates the ownership of `/etc/apache2` to `root:root` and we ended with apache not being able to read the ckan site directory. So it's important files are created with the correct permissions initially (`0644` or group `www-data`).
- The hardening playbook disables apache2, so the roles that depend on it will not be running correctly on the next reboot. This exacerbated by the umask 027, so if the service wasn't started at startup, `/run/<service>` is root:root and www-data can't read/write to it. If the service is started at boot time like normal, the umask hasn't kicked in yet, so all the `/run` dirs are created world readable, which is good.

I created a [branch](https://github.com/GSA/datagov-deploy/tree/hardening-permissions) with some fixes. Should I just push those here or create a separate PR?

JJediny:
Closed and merged into #406

JJediny:
> @hareeshreddyg and I tested this out today and ran into a few issues.
> 
> * The hardening playbook updates the ownership of `/etc/apache2` to `root:root` and we ended with apache not being able to read the ckan site directory. So it's important files are created with the correct permissions initially (`0644` or group `www-data`).
> * The hardening playbook disables apache2, so the roles that depend on it will not be running correctly on the next reboot. This exacerbated by the umask 027, so if the service wasn't started at startup, `/run/<service>` is root:root and www-data can't read/write to it. If the service is started at boot time like normal, the umask hasn't kicked in yet, so all the `/run` dirs are created world readable, which is good.
> 
> I created a [branch](https://github.com/GSA/datagov-deploy/tree/hardening-permissions) with some fixes. Should I just push those here or create a separate PR?

Addressed https://github.com/GSA/datagov-deploy/pull/407

PR #416, opened 2018-09-21
Add jekyll playbook to kitchen

adborden:


BangkitSedar (approved):

BangkitSedar (approved):

adborden:
The test is failing because supervisor is not starting properly. The role installs supervisor via pip instead of apt and then copies an old init-style script to /etc/init, so an ansible `service` start/restart won't do anything. Debating if we should re-work it to install supervisor from apt or if we should pull in a third-party supervisor role.

adborden:
@JJediny this is ready for review.

JJediny (approved):
This looks good

PR #450, opened 2018-10-24
Include redis password for celery config

adborden:
Seeing the fetch harvesters are failing to connect to redis with a NOAUTH error.
Celery needs to be configured with the redis password, too.


JJediny (approved):
Looks good

PR #463, opened 2018-11-06
[WIP] docker-build work

adborden:
Experimenting building docker containers using our ansible playbooks. In order to slim down the time of our CI builds, we can pre-build some base images for use with kitchen/molecule so we don't have to run the base playbooks.

This also makes it possible to pre-bake some scenarios rather than have to run an entire playbook before running our situational playbooks.

This works using a minimal ansible-ready container based on our platforms (Ubuntu 14.04) and the ansible docker connection. We create a special docker inventory where we can configure ansible variables specific to the containers.


adborden:
Note: I accidentally pushed these to `develop` and then reverted them. That's why the history is a little wonky. I re-configured our branch protection so admins can't push directly to the branch (admins can always disable this check if they really need to push something).

adborden:
lol https://github.com/ansible/ansible-container

adborden:
Closing as we're not going to go in this direction anytime soon.

PR #475, opened 2018-11-08
Jumpbox role

adborden:
Updates the jumpbox role, makes it more compatible for use with packer.

I think I'm going to pull this out into it's own repo after this is merged.


adborden:
@JJediny is the `AllowUser` directive in the SSH config required for compliance? I think I need to tweak this to allow for non-operator users like the secops user and the ubuntu user.

JJediny (changes requested):
We need to add a handler to restart openssh-server after config changes

adborden:
@JJediny did you want to add your config? https://github.com/GSA/datagov-deploy/blob/b733e5e34311155b1a1d33ac29008b086de9fd2f/ansible/group_vars/all.yml#L38

adborden:
Based on the convo with @JJediny , the ssh AllowUser config is not needed for compliance. I've removed it and there's no longer any config changes to ssh, so I removed the restart handler as well. This is ready for another look.

JJediny (approved):
Looks great ill add my public key as another PR

PR #476, opened 2018-11-08
[WIP] Packer jumpbox

adborden:
Adds packer template for jumpbox role

Some open questions:
- [ ] do we add this to CI, if so, whats the workflow?
- [ ] how do we cleanup old AMI images, or prevent them from growing in number?
- [ ] which VPC should we build images in? We've hit the limit on VPCs and can't create a new one dedicated to packer/testing use.


JJediny (approved):

adborden:
This isn't needed for now.

PR #478, opened 2018-11-09
Redirect requests containing %0A or %0D to 403.

carwin:


ansible/roles/software/wordpress/datagov-sudo-2-init/templates/datagov.nginx.conf
```diff
@@ -7,6 +7,7 @@ upstream php {
 server {
     listen 80 default_server;
     server_name _;
+    if ( $request_uri ~* "%0A|%0D" ) { return 403; }
```
adborden, line 10: 👍 should this also be applied to the https config?

carwin: Yes, that probably makes sense though I don't recall if https was specifically called out in the h1 issue.

adborden (approved):

PR #479, opened 2018-11-14
Run CI in BSP

adborden:
This adds a CI role for the jumpbox. It configures the jumpbox with a CI user that runs some of the playbooks in `--check` mode regularly by cron. After running the playbook, there is some manual setup for ansible vault and to configure the SSH key.

- CI job only contains the CI, catalog-web, and harvester playbooks
- job runs every 6 hours


JJediny (approved):
Looks good, we prob want to add skip tags as a bash variable in the future

PR #485, opened 2018-11-14
PR for unattended upgrades

JJediny:
@adborden looking for comments or feedback


ansible/unattended-upgrades.yml
```diff
@@ -0,0 +1,10 @@
+- hosts: all
+  roles:
+  - role: jnv.unattended-upgrades
+    unattended_origins_patterns:
+    - 'origin=Ubuntu,archive=${distro_codename}-security'
+    - 'o=Ubuntu,a=${distro_codename}'
+    - 'o=Ubuntu,a=${distro_codename}-updates'
+    - 'o=Ubuntu,a=${distro_codename}-proposed-updates'
```
adborden: Aren't the proposed updates considered less stable? In that case we wouldn't want to pull from those, right?

adborden (approved):

PR #489, opened 2018-11-15
[newrelic] install python newrelic to virtualenv

adborden:
This is a bit of a hack, since I'm hard-coding the virtualenv path in the role. Ultimately I think this playbook will go away, so I'm okay with this shortcut for now.

Newrelic should be integrated into catalog-app or the inventory role, since it requires installing python dependencies into the virtualenv. The configuration can be turned on/off through the config or environment variable based on the environment but it really should be considered an application dependency, not an os-level concern (and not a datagov-deploy concern)


JJediny (approved):
OK

PR #490, opened 2018-11-16
[common] set umask before installing python

adborden:
The hardening playbook sets a global umask 027. If the python install happens
after the hardening playbook, python files will be installed where they are not
world-readable, meaning ckan will fail (among other things).


JJediny (approved):
ok

PR #502, opened 2018-11-20
Set group_vars for the nginx role to use official nginx ppa.

carwin:


adborden (commented):
:+1: The file name is actually incorrect. It should be `ansible/group_vars/all/vars.yml`. I think these settings are not actually being applied because of this.

carwin:
@adborden are you sure? I ran the datagov-web play in my VM and I got the correct version when I added the vars here. Would you like me to try renaming the file and testing again?

adborden:
I'm sorry, I was mistaken. Just tested this out with `ansible-inventory` and it looks good.

adborden (approved):

PR #530, opened 2018-12-04
Remove terraform code

adborden:
This is no longer needed. https://github.com/GSA/datagov-infrastructure-live can
be used to create a testing environment.


JJediny (approved):
:+1:

PR #534, opened 2018-12-07
remove old and unused files

JJediny:
@adborden missing any?


.gitignore
```diff
@@ -57,5 +57,6 @@ ansible/inventories/mgmt-ocsit
 
 # Kitchen
 .kitchen
+Gemfile.lock
```
adborden: I don't think we want to ignore this. It helps keep versions consistent between developers and CI (since we're only using ruby dependencies for kitchen).

adborden (approved):

adborden:
There's a few more top-level files I think we can remove:
- https://github.com/GSA/datagov-deploy/blob/develop/staging.yml
- https://github.com/GSA/datagov-deploy/blob/develop/inventory
- https://github.com/GSA/datagov-deploy/blob/develop/DEPENDENCIES.md

.gitignore
JJediny: ok... restored

PR #555, opened 2018-12-15
Run plays on hosts in serial

adborden:
When running playbooks on instances, we only want to affect a few at a time to avoid downtime. We have so few hosts, that running them one at a time is the only reasonable option.

I didn't go through this with a fine-tooth comb, so I'm sure there are places we can optimize. I just wanted to start at the most conservative option. We can also set serial as a variable `serial: "{{ datagov_serial }}"` so there are different settings per environment. e.g. in staging we don't really need a serial option.

This also enabled ansible retry files so that it's easier to continue where you left off if a playbook fails.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -245,7 +245,6 @@ nocows = 1
 # You can disable this feature by setting retry_files_enabled to False
 # and you can change the location of the files by setting retry_files_save_path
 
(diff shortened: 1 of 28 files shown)


JJediny (approved):

adborden:
Rebased to pick up Jekyll fix.

PR #560, opened 2018-12-19
[ckan] remove Directory directive for apache

adborden:
This resolves nesuss 10704, Apache responding to MultiViews requests.

I'm not sure why this directive was ever included. It's not mentioned in the [CKAN installation docs](https://docs.ckan.org/en/ckan-2.3.5/maintaining/installing/deployment.html#create-the-apache-config-file).


adborden:
@JJediny can you take a look at this? It resolves one of the Apache items for Nessus.

JJediny (changes requested):
Have we first tested this by removing the configuration and restarting dev catalog instances, It was along time ago but i remember there was a reason for this, perhaps no longer though

adborden:
Hmm, yeah I misdiagnosed this. The Indexes are being served for /var/www which is in the apache2.conf. I'll submit a new PR.

PR #561, opened 2018-12-19
Include SCRIPT_FILENAME for php fpm

adborden:
Since we upgraded nginx, this is no longer included in fastcgi_params, so we
include it in our site config instead.

This fixes the php apps showing a blank page.


JJediny (approved):
LG

PR #564, opened 2018-12-19
Specify become: no on some plays

adborden:
These plays depend on the ssh user and aren't supposed to be run as root.
Instead, the plays should run individual tasks with the correct user instead of
relying on the ssh user.

This is just a quick fix to avoid a permissions issue with the nginx apps.


JJediny (approved):
ok

PR #568, opened 2019-02-04
Script to generate CSR from ansible-inventory

adborden:
In order to create GSA signed SSL/TLS certificates for our hosts, we use this
script to generate the CSR requests (one per inventory) to submit to GSA.


JJediny (approved):
Looks great

PR #569, opened 2019-02-05
Check reboot-required before rebooting

adborden:
This adds a check for `/var/run/reboot-required` and only reboots the host if the file exists. You can specify `ansible-playbook -e force_reboot=yes` to force a reboot.

I also tweaked the `wait_for` task. It seemed to be waiting for the full 300 second timeout even if the host came back up.


JJediny (approved):
LGTM

PR #575, opened 2019-02-08
Jekyll supervisor

adborden:
Updates how we refresh supervisor when updating the jekyll configuration. We introduced a new jekyll site and supervisor wasn't actually updating with the new site config without some manual help.


JJediny (approved):
:+1:

PR #577, opened 2019-02-08
Remove unused inventories

adborden:


JJediny (approved):
Confirmed no files are still in use LGTM

PR #583, opened 2019-02-11
Split tests on CI differently

adborden:
This refactors the way we run tests in parallel on CI. Before, we created a separate CI job for each suite. This is cumbersome and error prone because when you add a new suite, you have to add it in 3 places (once to Makefile, twice to CircleCI config).

This instead uses CircleCI's split tests functionality, we list out the individual tests and CI figures out how to split them across nodes. This means we only need to our new suites to the Makefile and then we're done.


JJediny (approved):
Looks great

PR #584, opened 2019-02-12
Common playbook

adborden:
Create's a "common" playbook that is applied to all hosts which contains all the smaller playbooks/roles. This makes the site.yml playbook a little cleaner and removes those common roles out of the application specific playbooks (which makes them quicker to test and apply).


JJediny (approved):
LGTM

PR #585, opened 2019-02-13
TLS role

adborden:
Adds host certificates, signed by GSA. Each environment has it's own certificate that covers all hosts.

This also includes some fixes for apache2 related to the Nessus scans.


JJediny (approved):
:+1:

adborden:
Rebased to pull in CI changes. Added a test key/certificate so apache2 starts properly.

PR #587, opened 2019-02-15
adding template for secops filebeat installer

JJediny:
@adborden this is just the template for testing, full vault update and prod config will be made as a new PR


ansible/roles/vendor/gsa.ansible-role-beats/molecule/default/create.yml
```diff
@@ -0,0 +1,81 @@
+---
```
adborden, line 1: Usually you don't need to include any molecule playbooks except for `playbook.yml`, molecule has some default create, destroy, etc playbooks baked in that it will use.

Are these modified from the default?

ansible/roles/vendor/gsa.ansible-role-beats/templates/config.j2
```diff
@@ -0,0 +1 @@
+{{ package_config | to_yaml }}
```
adborden, line 1: BTW, instead of using `template` for simple content, you can use [copy](https://docs.ansible.com/ansible/latest/modules/copy_module.html) with a `content` attribute:

```yaml
- name: copy file
  copy:
    dest: "{{ filepath}}"
    content: "{{ package_config | to_yaml }}"
```

adborden (approved):
Looks great! I left a few minor comments related to the upstream role.

ansible/roles/vendor/gsa.ansible-role-beats/templates/config.j2
JJediny: Will track this in the repo for the role

PR #589, opened 2019-02-16
[ckan] fix solr host for native-login

adborden:
The ckan-native-login playbook is usually run for the test environment where we don't have saml authentication, but we plan on using it in staging/production as part of the dedupe setup.


JJediny (approved):
:+1:

adborden:
Rebased to pull in CI changes.

PR #609, opened 2019-03-04
Newrelic attributes

adborden:
Set some attributes for newrelic infastructure so we can be more specific about targeting alerts for production vs staging vs test.


JJediny (approved):
Looks fine

PR #610, opened 2019-03-04
[WIP] Site playbook

adborden:
This removes the `site.sh` script and puts the remaining "app" playbooks into `site.yml`. This lets us deploy the entire data.gov platform from a single playbook, which paves the way for a clean CI/CD test. It also makes deploying much simpler. One command, no tags.

I cleaned up the tags in the plays that required tags so that by default, you get a standard provision without specifying tags. I also added documentation for the tags that still exist. The philosophy is that by default, you get everything, so you are sure the site is configured properly. As an option, you could specify tags only to update certain components (apache2, nginx, etc) or the app deploy.

There's still a lot of clean up to do. I left the database roles, but they are not used by anything. I believe these roles should be moved into the app code itself e.g. migrations should be run on app startup.

Eventually, I could see a CI/CD task watching datagov-deploy that runs the site playbook. Also, tasks watching individual app repos that run the appropriate playbook with the deploy tag.


adborden:
Spoke with @JJediny , let's remove the common playbooks from the Makefile, and only run the common playbook on CI. Eventually these roles will be broken out into their own repo. For now, having the many small builds run on every commit is overwhelming.

adborden:
Closing

PR #616, opened 2019-03-05
[common] reboot notifications

adborden:
- Configure unattended-upgrades to only email on error
- Add a cron job to notify when a reboot is required


JJediny (approved):
👍

PR #624, opened 2019-03-06
Remove unused role

adborden:
software/catalog/nginx isn't used anywhere.


adborden:
Oops, I guess it's used in web-proxy.yml. @JJediny is that used for anything? I don't think it is.

JJediny (changes requested):
The nginx proxy is used to proxy pycsw to `/csw` on catalog.data.gov it was never fully implemented

adborden:
Let's dig into this a little more. I think apache is handling the [proxying to pycsw](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/catalog/ckan-app/templates/etc/apache2/sites-enabled/ckan.default.conf.j2#L14). It looks like nginx isn't installed on the production catalog-web hosts.

adborden:
Discussed offline, Apache is currently handling the proxy. Nginx was never implemented, so we're going to go ahead and remove web-proxy.yml.

JJediny (approved):

adborden:
Removed web-proxy playbook.

PR #635, opened 2019-03-08
Install GSA root CA

adborden:
Installs the GSA root CA on our hosts as well as adding molecule tests. The GSA root CA is needed in order for GSA internal certificates, like our host certificates or catalog-datagov.dev-ocsit.bsp.gsa.gov. Without this, most maintenance jobs fail in staging because talking to the catalog API fails certificate validation.

PR #636, opened 2019-03-09
Common packages

adborden:
Depends on https://github.com/GSA/datagov-deploy/pull/635

Provide a cron service and a service manager as part of the platform on every host. These are probably already installed, but we want to make it explicit. For the service manager, we're using supervisor for trusty, systemd for bionic.


JJediny (approved):
👍

PR #638, opened 2019-03-11
[WIP] Swap logging stack from EFK to ELK/filebeats

JJediny:


JJediny:
Closing until we're able to revisit swapping out trendmicro for ossec-hids, for now we are reverting to using filebeats for secops logging requirements and dedicated fluentd for application logging stack

Issue #574, opened 2019-02-07
Renew site certificates


adborden:
The `*.data.gov` certificate expires April 21, 2019. GSA offers external certificates. We should look into this as an option.

hkdctol:
We will create Service Catalog request to request the external certs from GSA. For that we need the complete list of subdomains. Will send to @adborden

adborden:
Service Now ticket submitted.

adborden:
This got hung up related to DNSSEC policy. We're waiting on GSA cert team to clarify the requirements. In the meantime, we're looking for alternatives.

adborden:
:tada: the certificate has been issued. Key and certificate are in [drive](https://drive.google.com/drive/folders/1vMCI5YXvVklSZTEohfnZke04upo8qdJc). I'll open a ticket with BSP to get these installed on NetScaler

@JJediny can you install these on the CDN?

adborden:
Certificate and key have been installed on Cloud Front. Opened [RITM0601967](https://gsa.service-now.com/sp/?sys_id=c2c1afb4db307f080e1d368d7c961920&view=sp&id=ticket&table=sc_req_item) to install to NetScaler.

adborden:
Verified that domains are using the new certificate with [this script](https://gist.github.com/adborden/f0dc1cfa0eb8912ec4def8322860377c). 

The following domains were not updated, but I believe they are no longer used or are manged by another team.
```
coop.smallbusiness.data.gov... fail
manage.data.gov... fail
origins-data-gov.data.gov... fail
origins-smallbusiness.data.gov... fail
smallbusiness.data.gov... fail
```

adborden:
For reference, this is the content of the ticket for BSP. Just like any other BSP ticket, it should be a Generic Request:

> Short Title: Replace *.data.gov site certificate
>
> Key and certificate are available in Google Drive, please request access https://drive.google.com/drive/folders/1vMCI5YXvVklSZTEohfnZke04upo8qdJc
> 
> There are about ~80 domains to address.

PR #646, opened 2019-03-15
Generate CSR for site certificates

adborden:
Include list of production SANs in version control.

Also updated https://github.com/GSA/datagov-deploy/wiki/TLS-SSL-Certificates

Part of #574


adborden:
Hmm, so the site certificates/keys aren't going to get committed to ansible inventory. Should we store them in Google Drive?

JJediny (approved):
Yes we can store them in google drive

adborden:
https://drive.google.com/drive/folders/1vMCI5YXvVklSZTEohfnZke04upo8qdJc

PR #653, opened 2019-03-19
Harvester supervisor cleanup

adborden:
- adds molecule to catalog-harvester
- adds bionic support
- moves to conf.d style supervisor configs
- reduces number of worker processes
- uses supervisor for all cron jobs


adborden:
@JJediny can you take a look at this?

JJediny (approved):
Looks good

PR #664, opened 2019-03-22
Fix elastic kibana deploy

JJediny:
Ready for review


ansible/kibana.yml
```diff
@@ -7,3 +7,5 @@
 
   roles:
     - role: gsa.ansible-role-kibana
+      vars:
+        kibana_version: "6.6"
```
adborden, line 11: Should this kind of variable be in the playbook or in the inventory?

ansible/inventories/mgmt/group_vars/all/vars.yml
```diff
@@ -1,3 +1,17 @@
...
+# Elasticsearch/Kibana shared creds
+kibana_elasticsearch_username: "{{ vault_kibana_elasticsearch_username }}"
+kibana_elasticsearch_password: "{{ vault_kibana_elasticsearch_password }}"
+elasticsearch_name: "{{ vault_elasticsearch_name }}"
+elasticsearch_port: 9200
+kibana_port: 5601
+
+# Nginx basic auth
+basic_auth_user: "{{ vault_basic_auth_user }}"
+basic_auth_hash: "{{ vault_basic_auth_hash }}"
```
adborden, line 14: I didn't see these in the vault, maybe they are missing the `vault_` prefix in the vault?

adborden (approved):

ansible/kibana.yml
JJediny: I moved all other vars to inventory, but I wanted to make this declarative in the playbook to make it easier to upgrade and make it more obvious for the user deploying the role.

ansible/inventories/mgmt/group_vars/all/vars.yml
JJediny: Good catch, added a commit to prepend the missing `vault_`

Issue #622, opened 2019-03-05
software/catalog/security-fix should be moved to catalog-app


adborden:
[software/catalog/security-fix role](https://github.com/GSA/datagov-deploy/blob/master/ansible/roles/software/ckan/security-fix/tasks/main.yml) includes version bumps that should be managed by catalog-app and the requirements.txt

adborden:
FIxed in https://github.com/GSA/datagov-deploy/pull/684

PR #684, opened 2019-03-29
Remove software/ckan/security-fix

adborden:
Dependencies must be managed by the application, not the platform. These
versions are already outdated and are not being checked by our dependency scans.


adborden:
Fixes https://github.com/GSA/datagov-deploy/issues/622

JJediny (approved):
👍

Issue #640, opened 2019-03-11
Specify dev branch for jekyll sites on staging servers


philipashlock:
The Jekyll configuration is currently defaulting to the master branch when deploying staging sites, but ideally we would be able to specify an alternate staging/dev branch and config file for those instances. This should be specified in jekyll.yml

* For `data-strategy` this can be set to the `develop` branch and `_config_staging.yml` for the config file

I'm not sure we properly have staging sites set up for the others, but anyway:
* For `data-federation` this can be set to `staging`
* For `sdg-indicators-usa` this can be the same as prod, `prod-pages` for now

Relevant config files for reference:
`ansible/jekyll.yml`
`ansible/roles/software/jekyll/templates/etc/supervisor/conf.d/jekyll.conf.j2`

PR #685, opened 2019-03-29
[jekyll] specify separate staging config for apps

adborden:
Resolves https://github.com/GSA/datagov-deploy/issues/640


JJediny (approved):
Moving these sites to federalist but no harm LGTM

PR #687, opened 2019-03-31
Refactor catalog plays for site playbook

adborden:
- Splits catalog-web and catalog-harvester plays
- Removes unused roles


adborden:
This replaces the catalog work in https://github.com/GSA/datagov-deploy/pull/610

ansible/site.yml
```diff
@@ -10,3 +10,8 @@
 - import_playbook: nessus.yml
 - import_playbook: newrelic-infrastructure.yml
 - import_playbook: unattended-upgrades.yml
+
+
+# App plays
+############
+- import_playbook: catalog.yml
```
JJediny, line 17: Was this meant to be included? I'm deleting the site.yml playbook altogether with the refactoring work

JJediny (commented):
Looks good but can you change the ckan-worker to ckan-harvest as there is another kind of ckan instance that is the qa worker and I don't want to be mixing vocabs

JJediny (approved):
LGTM

PR #702, opened 2019-04-04
Add Pipfile

adborden:
Adds a Pipfile, but leaves requirements.txt while we transition away from Kitchen. This avoids having to deal with a handful of issues around CI, the machine executor, and kitchen.


JJediny (approved):
lg

Issue #402, opened 2018-09-10
Refactor solr plays into it's own repo/role with tests


adborden:

adborden:
Done in https://github.com/GSA/datagov-deploy/pull/735

Issue #371, opened 2018-07-26
Grant tomcat permissions to /opt/newrelic/logs with ansible


JJediny:
solr shutdown unexpectedly because of a permissions error on a dir owned by root user which tomcat was trying to write to logs 

solution to be implemented in ansible:
`chown -R tomcat6:tomcat6 /opt/newrelic/ ` resolved

adborden:
N/A for the updated solr instances.

PR #735, opened 2019-04-10
Solr for bionic hosts

adborden:
- Removes the old non-idempotent solr role.
- Moves the old `solr` hosts to `solr4`, so they won't be matched by the solr playbook.
- Adds solr playbook to the site playbook.


JJediny (approved):
LG

adborden:
Resolves #402 #371

PR #761, opened 2019-04-26
[WIP] Feature/switch php apps to molecule

JJediny:
Holistic refactoring of php apps

TODO:
- [ ] validate dir structure with nginx conf with current setup 
- [ ] ensure mysql init and migration works
- [ ] get wordpress under CI - which has yet to be done
- [ ] apply and validate roles in sandbox


JJediny:
Only useful for reference so closing

PR #787, opened 2019-05-10
[catalog] reload supervisor on deploy

adborden:
When new code is deployed, harvester processes need to be restarted.


JJediny (approved):
LGTM

PR #806, opened 2019-05-29
[inventory] update who.timeout to match beaker

adborden:
Hoping to solve the issue of random logouts when browsing inventory pages.


JJediny (approved):
ok

Issue #713, opened 2019-04-08
Ubuntu Bionic rollout to solr host(s)


adborden:
## User Story
As an operator, I need new Ubuntu Bionic hosts provisioned in all of our environments so that I can continue receiving security updates automatically from Ubuntu.

Part of https://github.com/GSA/datagov-deploy/issues/571

Ubuntu 14.04 support ends [April 2019](https://www.ubuntu.com/about/release-cycle). We should move to Ubuntu 16.04 or 18.04. The latest 18.04 is preferred, but we should consider what kind of support exists for it.

## Details
Hosts are provisioned in these environments and play books are successfully applied.
1. [x] bionic sandbox
1. [x] staging
1. [x] production

To rebuild the solr index:
1. [x] configure a single harvest instance to use solr, disable any cron jobs
1. [x] rebuild the index `ckan seach-index rebuild_fast`
1. [x] update catalog inventory and redeploy so all catalog instances are using the new solr
1. [x] do the same for inventory

## Acceptance Criteria
- [x] Bionic hosts are provisioned and operating correctly
- [x] Bionic hosts reporting to our monitoring systems
- [x] Solr index is rebuilt
- [x] Old Trusty hosts are decommissioned and removed from Ansible inventory
- [x] New Relic solr alarm is updated for the new process name

## Definition of Done

The above acceptance criteria must be met.

adborden:
[BSP ticket](https://gsa.service-now.com/sp/?id=form&table=sc_req_item&filter=&sys_id=298f57f7dbe477c40e1d368d7c961943&v=) opened, waiting for them to provision hosts.

adborden:
Hosts were provisioned, but with incorrect EBS volumes configured. Waiting for them to be re-provisioned.

adborden:
EBS volumes have been attached to the instance and I initialized them. I've added steps for rebuilding the solr index and I'm currently testing them on staging.

adborden:
Waiting on SecOps to provide a trendmicro package for Bionic.

adborden:
Trendmicro issue is resolved

I noticed that the solr replicas aren't sized correctly. They are t2.medium but should be m4.xlarge.

adborden:
Inventory index is rebuilt in staging and production. Inventory hosts are now running against these new solr instances 🎉

adborden:
[...]

PR #818, opened 2019-06-14
Hotfix: use bionic solr in production (master)

adborden:
Part of https://github.com/GSA/datagov-deploy/issues/713


JJediny (approved):
LG

PR #820, opened 2019-06-14
Fixes for NewRelic python

adborden:
- Add missing license key
- Specify an application name
- Remove `current_env` which doesn't work
- Adds catalog-admin inventory group to create admin as a separate NewRelic application


JJediny (approved):
LG

PR #827, opened 2019-06-20
Remove trusty hosts

adborden:


JJediny (approved):
LG but should we comment out until those instances are deleted? I'm fine with as is

PR #847, opened 2019-07-12
Remove requirements.txt

adborden:
This avoids dependabot trying to update requirements.txt which is redundant with Pipfile.lock.


JJediny (approved):
Looks good to go

Issue #488, opened 2018-11-15
Disable CKAN tracking


adborden:
I'm seeing this error in the logs on catalog-web2d:
```
[Thu Nov 15 01:26:23.586763 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095] mod_wsgi (pid=1789): Exception occurred processing WSGI script '/etc/ckan/apache.wsgi'.
[Thu Nov 15 01:26:23.586901 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095] Traceback (most recent call last):
[Thu Nov 15 01:26:23.586934 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]   File "/usr/lib/ckan-new/src/ckan/ckan/config/middleware.py", line 360, in __call__
[Thu Nov 15 01:26:23.586977 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]     self.engine.execute(sql, key, data.get('url'), data.get('type'))
[Thu Nov 15 01:26:23.586993 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]   File "/usr/lib/ckan/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 1690, in execute
[Thu Nov 15 01:26:23.587020 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]     return connection.execute(statement, *multiparams, **params)
[Thu Nov 15 01:26:23.587032 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]   File "/usr/lib/ckan/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 719, in execute
[Thu Nov 15 01:26:23.587049 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]     return self._execute_text(object, multiparams, params)
[Thu Nov 15 01:26:23.587061 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]   File "/usr/lib/ckan/lib/python2.7/site-packages/sqlalchemy/engine/base.py", line 868, in _execute_text
[Thu Nov 15 01:26:23.587077 2018] [:error] [pid 1789:tid 139755571001088] [remote 10.177.19.22:21095]     statement, parameters
[...]

PR #873, opened 2019-08-05
Disable tracking

adborden:
Reduce the number of 500s we're seeing until we can resolve.
https://github.com/GSA/datagov-deploy/issues/488


adborden:
Resolves #488

PR #912, opened 2019-08-23
Include logrotate in common playbook

adborden:
This should be applied to all hosts, so let's include it in the common playbook.


pburkholder (approved):
I approve, but I can't vouch for understanding the impliacations of this review. P

PR #928, opened 2019-08-29
[logrotate] rotate weekly

adborden:
Rotating weekly is a little easier for log auditing. If we review the logs
weekly, we only have a single log file to inspect.

Remove the size option, which overrides the daily/weekly directive.

Remove the duplicate apache config, since apache2 already exists from the
package.


pburkholder (approved):
LGTM

PR #932, opened 2019-09-04
[logrotate] remove configuration for OS logs

adborden:
We don't want to conflict with OS provided configuration. The maintainers know
better about how to rotate their own logs.


adborden:
Confirmed that our logrotate configuration is causing segfaults in apache2 when rotating the script
https://docs.google.com/document/d/1E9tcoU5nySs4MumqZkSbIKZrV5t7ecQ8tMeu0z8ogNk/edit#

pburkholder (approved):

adborden:
😔 the log disks on catslog-web are filling up with the OS provided config. The Apache config is weekly rotation with 52 logs, a whole year of data. I guess we'll want to be editing the Apache config after all.

adborden:
... or we move the ckan logs to another directory so we can use a custom config without affecting Apache.

adborden:
Rebased to pickup CI fix.

PR #933, opened 2019-09-04
Use templating only on .env and use php from app repo

pburkholder:


ansible/roles/software/dashboard/dashboard-deploy/templates/.env
```diff
@@ -6,3 +6,12 @@ DB_USER='{{ env_db_user }}'
 ENCRYPTION_KEY='{{ env_encryption_key }}'
 
 PRE_APPROVED_ADMINS='{{ codeigniter_pre_approved_admins | join(",") }}'
+
+S3_BUCKET=foobar
+S3_PREFIX=prefix
+USE_LOCAL_STORAGE=true
+
+PROTOCOL=http
+PROJECT_SHARED_PATH=/var/www/app
```
adborden: You'll want to template this since we install the app to a new directory on each deploy.

```suggestion
PROJECT_SHARED_PATH={{ current_source_symlink }}
```

pburkholder:
So the `converge` issues I was having are related to the settings for `database.php` configuration settings for `$db['default']['db_debug'] `. 

In `datagov-deploy` the setting is, at https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/dashboard/dashboard-deploy/templates/config/database.php#L71, 

```
$db['default']['db_debug'] = FALSE;
```

and in the pod-dashboard project we have it set as

```
$db['default']['db_debug'] = (ENVIRONMENT == 'development') ? TRUE : FALSE;
```

so the TRUE value seems to mean 'fail on db errors' and the FALSE 'continue on db errors and pretend everything is ok' :) 

Arguably, in non-development envs we'd want database errors to cause everything to come to a crashing halt, but we've been running with things as "FALSE", which also hides DB errors from teh end user.

PR #935, opened 2019-09-07
Fix CI

adborden:
Hitting the 10m timeout on CI with no output. I tried slimming down some of these install steps which take a while (download + install).

- swap jdk for jre, smaller download
- split some of the package installs into two steps


pburkholder (approved):
Makes sense to me.

PR #941, opened 2019-09-11
Hotfix: [jumpbox] add Phil

adborden:

PR #942, opened 2019-09-11
Hotfixes (develop)

adborden:
I was expecting to pull in https://github.com/GSA/datagov-deploy/pull/941 hotfix into develop, but looks like there were a few other hotfixes that haven't been pulled in.


pburkholder (approved):
One formatting comment, otherwise good to go

PR #948, opened 2019-09-12
Add sender_canonical_map to rewrite sender address for DMARC policy

adborden:
This updates the postfix configuration to map to
1. rewrite the sender address acceptable to DMARC so pretty much any mail is delivered
1. map root to the tech help email address

In effect, any email notification from all hosts will go to the tech team unless explicitly addressed otherwise.


pburkholder (approved):
Looks OK. I'm so sorry you have to touch postfix. :(

PR #956, opened 2019-09-14
Update smoke tests

adborden:
- Add a service-level smoke test (at the public url)
- Move the host-level tests closer to deploy to  prevent a bad deploy earlier
- Use more specific URLs that better indicate the health of the service
- Don't follow redirects in assertions so that we know exactly what we're testing

### DIFF
--- a/ansible/catalog-web.yml
+++ b/ansible/catalog-web.yml
@@ -30,7 +30,6 @@
       when: datagov_enable_postgresql_role is defined
 
 
-
 - name: Catalog web stack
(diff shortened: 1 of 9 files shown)


pburkholder (approved):
At some point perhaps quote-enclose all URLs, but looks good to roll.

adborden:
Interesting, I've never had an issue with unquoted urls. I prefer to only use quotes in yaml when necessary or to avoid ambiguity.

PR #960, opened 2019-09-20
Update logging configuration for CKAN

### LINKED ISSUE #949: /var/log filling up on catalog-web
The /var/log partition is filling up and triggering the 94% used alarm. audit and apache2 are the largest directories. ckan.error.log* are the largest files.

I think we want to move the CKAN logs to their own directory and add a logrotate config that better handles CKAN.

We should also check the log configuration. There seem to be a lot of INFO messages from `who` which don't seem all that useful.

adborden:
Fixes https://github.com/GSA/datagov-deploy/issues/949

Moves the CKAN logs to their own app directory. Adds an aggressive logrotate configuration for catalog.

### DIFF
--- a/ansible/roles/software/ckan/inventory/tasks/install.yml
+++ b/ansible/roles/software/ckan/inventory/tasks/install.yml
@@ -6,6 +6,9 @@
     - swig
     - xmlsec1
 
+- name: Create log directory
+  file: path={{ inventory_log_dir }} state=directory owner=root group=www-data mode=0750
+
 - name: Remove old rollback code
   file: path="{{ project_source_rollback_path }}" state=absent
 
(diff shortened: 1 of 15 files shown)


ansible/roles/software/ckan/inventory/tasks/install.yml
```diff
@@ -56,14 +59,15 @@
     - etc/ckan/who.ini
     - etc/ckan/apache.wsgi
     - etc/ckan/datapusher.wsgi
-    - etc/apache2/sites-enabled/ckan.conf
-    - etc/apache2/sites-enabled/datapusher.conf
 
 - name: Configure production.ini
-  template: src=templates/{{ item }} dest=/{{ item }} owner=root group=www-data mode=0640
+  template: src={{ item }} dest=/{{ item }} owner=root group=www-data mode=0640
```
pburkholder, line 64: Can you confirm that the `templates` wasn't working before because the `src=` was wrong?

pburkholder (approved):
One point I was curious about regarding `src=` for templates. But go ahead and merge when you're ready.

ansible/roles/software/ckan/inventory/tasks/install.yml
adborden: Ansible does some path magic ([checking different locations for files](https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html?highlight=templates#using-roles)), so this was working correctly before. `templates` is the default (or the first location to check) for the `template` module, so this is essentially removing redundant path information, and leveraging the well-known default role templates directory.

PR #966, opened 2019-09-23
Refactor Operators

adborden:
- Explicitly set operators as active
- Move operators out of inventory variables, hoist up to playbook variable in
  two groups that each environment can leverage
- Add common_operators for https://github.com/GSA/datagov-deploy-common/pull/28


ansible/roles/vendor/requirements.yml
```diff
@@ -59,7 +59,7 @@
   version: v2.2.3
 - name: gsa.datagov-deploy-jumpbox
   src: https://github.com/GSA/datagov-deploy-jumpbox
-  version: v1.1.1
+  version: v2.0.0
```
pburkholder, line 62: Just curious - where does ansible versioning happen? I thought it might be in gsa.datagov-deploy-jumpbox/meta/main.yml, but seems not.

pburkholder (approved):

ansible/roles/vendor/requirements.yml
adborden: It's by GitHub tag, unless published to https://galaxy.ansible.com (which we don't do).

Ansible role versioning is not great, but usable.

PR #969, opened 2019-09-24
[inventory] remove datapusher https hack

adborden:
I believe this was introduced because of the host certificates, which didn't
work without some extra CA configuration on staging (and maybe production).
After removing https://github.com/GSA/datagov-deploy/pull/965, datapusher might
just push to the service-level name which would have a proper https
configuration.

Either way, this hack should go as it's causes confusion about how datapusher
works with inventory.

### DIFF
--- a/ansible/roles/software/ckan/inventory/tasks/additional-tasks.yml
+++ b/ansible/roles/software/ckan/inventory/tasks/additional-tasks.yml
@@ -31,13 +31,6 @@
     replace: 'application/x-www-form-urlencoded'
   notify: restart apache2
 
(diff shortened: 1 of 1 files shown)


pburkholder (commented):
Does the current `/usr/lib/datapusher/lib/python2.7/site-packages/ckanserviceprovider/web.py` need to re-rewritten to use https instead of http, or will this go out with a fresh release?

adborden:
Yes, on each deploy we start with a fresh copy, so removing that Ansible task is enough.

crysacarter (approved):

adborden:
Rebased to resolve conflict.

PR #970, opened 2019-09-24
Roadmap update

adborden:


adborden:
@pburkholder do you think it's worth keeping these milestones documented here, or should they be moved to epics?

pburkholder (approved):

pburkholder:
I think this high-level narrative is good, pending the opinions of any incoming product manager, so they don't need to be epics until you're ready to start work.

PR #973, opened 2019-09-25
Lower pod-dashboard max_remote_size to 127Mb from 500Mb

pburkholder:
`get_file_contents` fails with max_size 500MB (and 256Mb) but seems OK w/ 127MB


adborden (approved):

adborden:
Not sure why CI is failing for this, seems to be the same issue that was fixed in https://github.com/GSA/datagov-deploy/pull/935

PR #988, opened 2019-09-26
Pull in hotfixes to develop

adborden:
Kind of surprised to see so many changes. I guess some hotfixes didn't make it into develop.


ansible/roles/monitoring/newrelic/python-agent-ansible/defaults/main.yml
```diff
@@ -2,3 +2,4 @@
 nrinfragent_license_key: licesense-goes-here
 ckan_config_path: '/etc/ckan'
 newrelic_config_path: '/etc/newrelic'
+newrelic_environment: production
```
pburkholder: Do you want a default value of `production`? Maybe a `default` environment for systems that don't get overwritten by a more specific value from group_vars.

pburkholder (approved):
1 question, LGTM otherwise. Merge when ready.

ansible/roles/monitoring/newrelic/python-agent-ansible/defaults/main.yml
adborden: Yeah, in this case I'd prefer production being the default. Just in case someone forgets to set the environment, I would rather see things show up in the production environment instead of accidentally ignoring it because it's showing up in non-production.

We only cover NR licensing for production and staging so the only time we might make this mistake is when standing up a new environment (like cloud.gov) and these Ansible roles would not apply.

Also, I'd like to move this configuration into the application code instead of Ansible.

PR #1015, opened 2019-10-03
Add exit criteria to Kanban columns

adborden:
Content pulled and remixed from
https://github.com/18F/cg-product/blob/3f58e4f3aa95a03d77726da6d79f7c185bda0505/StoryLifecycle.md

- [ ] Once reviewed, I'll update the Kanban board columns.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,8 +118,8 @@ Work that we are planning on doing and will groom and schedule into a sprint.
 - Indicate the intended benefit and who the story is for in the "as a ..., I want
   ..., so that ..." form.
 - Acceptance criteria is defined.
-- If necessary, the story includes a security testing plan. For example, the
-  acceptance criteria include automated tests and alerts for unexpected behavior.
+- If necessary, the story includes a security testing plan and any tasks from
+  that plan are included as acceptance criteria.
 
 
 #### Sprint backlog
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -30,30 +30,96 @@ task “Capacity”
...
+Work that we are planning on doing and will groom and schedule into a sprint.
+
+
+##### Exit criteria
+
+- Indicate the intended benefit and who the story is for in the "as a ..., I want
+  ..., so that ..." form.
+- Acceptance criteria is defined.
+- If necessary, the story includes a security testing plan. For example, the
+  acceptance criteria include automated tests and alerts for unexpected behavior.
```
mogul: Security testing seems orthogonal to automated tests and alerts. Maybe split this into two items?

adborden: Hmm, what I'm trying to say is, "we've thought about the security implications of this feature and have included any tasks to address those implications in the acceptance criteria. For example, if we need to assert that a URL is accessible only by authorized users, then an acceptance criteria is added to add an automated test that will test that URL."

so... maybe.. this?

> If necessary, the story includes a security testing plan and any tasks from that plan are included as acceptance criteria.

mogul: Sure, that's clearer.

adborden: Updated, ready for another review.

mogul (approved):

PR #1040, opened 2019-10-09
Use upstream beats role

adborden:
- [ ] https://github.com/GSA/datagov-deploy-common/pull/31

This resolves the issue of having to use `--skip-tags filebeat`.


adborden:
Superceded by https://github.com/GSA/datagov-deploy/pull/1344

PR #1054, opened 2019-10-16
Be clearer about using git-flow patterns, not tools

mogul:


adborden (approved):
Much clearer, ecto-1.

Issue #805, opened 2019-05-24
Move supervisor services to systemd


adborden:
## User Story

As an operator, I want all the supervisor jobs/services running as systemd jobs/services so that I get the features of systemd, including failure notifications and resource limits.

## Details

An advantage to moving to Ubuntu Bionic is that we get systemd. We should update the supervisor jobs to take advantage of this.

## Acceptance Criteria

- [ ] Supervisor jobs are move to systemd
- [ ] Where appropriate, job failures send a notification to datagovhelp@
- [ ] Supervisor is uninstalled from all hosts

## Definition of Done
> Data.gov jobs and services run under systemd just like OS level jobs and services.

nickumia-reisys:
- https://github.com/GSA/data.gov/issues/3710

PR #1056, opened 2019-10-17
Fix dashboard cron jobs overrunning each other

### LINKED ISSUE #1050: Dashboard High Memory alerts from overlapping cron jobs
We keep seeing dashboard1p and dashboard1d go down or [alert on high memory](https://alerts.newrelic.com/accounts/1601367/incidents/89067741/violations). This appears to be cron jobs running long, and then new reruns overlap, consuming more memory without cleaning up the last one.

We can move to supervisor to prevent new runs starting when the previous one hasn't finished. [timeout](https://linux.die.net/man/1/timeout) is part of GNU coreutils and can kill the job after a timeout.

## How to reproduce

1. dashboard cfo-act download job runs daily. On some occasions, the runs don't end and don't clean up resources.

## Expected behavior

Jobs run one at a time. If a job runs too long, it is killed.

## Actual behavior

Jobs can run indefinitely, and be run on top of each other, consuming more and more memory.

adborden:
Fixes #1050 
- Adds molecule
- Moves cron jobs to supervisor
- Moves logging to directory
- Adds logrotate for dashboard cron jobs

This doesn't address any of the Dashboard web app logging issues.

### DIFF
--- a/ansible/roles/software/dashboard/dashboard-deploy/molecule/default/tests/test_default.py
+++ b/ansible/roles/software/dashboard/dashboard-deploy/molecule/default/tests/test_default.py
@@ -1,0 +1,12 @@
+import os
+
+import testinfra.utils.ansible_runner
+
+testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
+    os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
+
+
+def test_hosts_file(host):
+    hosts = host.file('/etc/hosts')
+
+    assert hosts.exists
--- a/ansible/roles/software/dashboard/dashboard-deploy/molecule/default/tests/test_default.py
+++ b/ansible/roles/software/dashboard/dashboard-deploy/molecule/default/tests/test_default.py
@@ -6,7 +6,36 @@ testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
     os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
 
 
-def test_hosts_file(host):
-    hosts = host.file('/etc/hosts')
+dashboard_user = 'ubuntu'
 
-    assert hosts.exists
+
+def test_log_dir(host):
+    log_dir = host.file('/var/log/dashboard')
+
+    assert log_dir.exists
+    assert log_dir.is_directory
+    assert log_dir.user == dashboard_user
+    assert log_dir.group == dashboard_user
+
+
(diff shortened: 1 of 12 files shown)


ansible/roles/software/dashboard/dashboard-deploy/molecule/default/tests/test_default.py
```diff
@@ -6,7 +6,36 @@
...
+
+def test_cron(host):
+    cron = host.file('/etc/cron.d/dashboard')
+
+    assert cron.exists
+    assert cron.user == 'root'
+    assert cron.group == 'root'
+    assert cron.mode == 0o644
+    assert cron.contains('supervisorctl start dashboard-cfo-act-download')
+    assert cron.contains('supervisorctl start dashboard-cfo-act-full-scan')
```
mogul, line 41: I'm a little confused here... So cron is still used to ensure the supervisor jobs are started? Isn't there a better mechanism for that, eg chkconfig? Or maybe the intent is just to recover in case one of them dies...?

mogul (commented):
I can understand the content of the two commits, but my low experience with Ansible and Supervisor means I don't have more substantive feedback than this... Can't say what's good/bad/wrong/could-be-better! But in any case, it seems like the right direction to take.

ansible/roles/software/dashboard/dashboard-deploy/molecule/default/tests/test_default.py
adborden: These jobs are not long-running processes, they are one-off tasks that exit 0 once the task is complete. cron schedules the job at a specific time and supervisor acts more like a task manager. Supervisor provides:
- task arguments "as code" since they're baked into the config
- parallelism of a task (or in this case, enforcing a singleton task)
- retries on failure (no retries in this case)
- logging

We use supervisor for both these one-off tasks and for long-running tasks/services.

Once we're completely on bionic, we'd probably [move to systemd](https://github.com/GSA/datagov-deploy/issues/805) which has more features like resource limits, watch dogs, and timers (dropping the need for cron).

mogul: That makes sense to me... I was looking for a `chkconfig` or something somewhere.

adborden:
@mogul can you take another look?

mogul (approved):
I don't see anything eyebrow-raising offhand, though my limited experience with Molecule, Upstart, and Ansible means I might be missing things.

Issue #919, opened 2019-08-26
Upgrade PHP applications to use PHP 7.3


adborden:
## User Story
As an operator, I want the PHP applications to run on PHP 7.3.x so that they have the latest security fixes and the longest supported lifetime.

## Details
https://www.php.net/supported-versions.php

This should include all the PHP applications:
- CRM
- Dashboard
- WordPress

## Acceptance Criteria
- [ ] Application development environments and tests are updated
- [ ] Deployment playbooks are updated

adborden:
For this task, you'll want to first test the PHP changes locally in these repositories:

- [CRM](https://github.com/GSA/open311-simple-crm) ([with docker-compose PR](https://github.com/GSA/open311-simple-crm/pull/38))
- [Dashboard](https://github.com/GSA/project-open-data-dashboard) (@pburkholder added docker-compose already)
- [WordPress](https://github.com/GSA/datagov-wp-boilerplate) ([with docker-compose PR](https://github.com/GSA/datagov-wp-boilerplate/pull/25))

Then, we'll want to update the Ansible playbooks which will be applied to the BSP environment.
- [CRM roles](https://github.com/GSA/datagov-deploy/tree/develop/ansible/roles/software/crm)
- [Dashboard roles](https://github.com/GSA/datagov-deploy/tree/develop/ansible/roles/software/dashboard)
- [WordPress roles](https://github.com/GSA/datagov-deploy/tree/develop/ansible/roles/software/wordpress)

adborden:
And it's probably to best to start with one of them (CRM) and take it all the way to production before moving on to the other apps.

mogul:
I'm working on updating the dashboard role for the BSP environment and getting a little lost on how to test it. Let's meet early Monday to review, @adborden!

mogul:
> And it's probably to best to start with one of them (CRM) and take it all the way to production before moving on to the other apps.

Per discussion today, it's easier to keep the common Ansible roles intact if we do them all at once, so I'm taking that path instead.

adborden:
@mogul deployed this to production today.

PR #1059, opened 2019-10-21
Upgrade to PHP 7.3.x

mogul:
Addresses #919 

Confirmed `dashboard` is working and passing tests with `bundle exec kitchen test dashboard`. 

CRM and WordPress still need to be tested as working before it can be deployed, hence the draft state.


adborden:
Awesome so far! One thing to add (maybe)... We disabled Kitchen in CI a while ago, and it would be great to re-enable it. Basically just reversing [this change to the Makefile](https://github.com/GSA/datagov-deploy/commit/b96874b9d68bea8e3bcb5e190dc80e5971b61718#diff-b67911656ef5d18c4ae36cb6741b7965). We would still need to resolve the vault issue... Still scratching my head on how to address that one.

mogul:
Flailed about a bit in `ci` but still failed to get a deployment going. I think I must be misunderstanding what needs to happen for my SSH key to be in the right place. Let's debug together Wednesday morning!

mogul:
Things noted along the way that are out of scope for this PR; should be separate issues: 

* [x] The initial .sql import for CRM should move into a migration, and be tested for idempotentcy --> https://github.com/GSA/datagov-deploy/issues/1098
* [x] We need to add the CI env config for dashboard (terraform + vault + vars) and verify that .env gets populated properly --> https://github.com/GSA/datagov-deploy/issues/1100

mogul:
Rebased (a few times 😬 ) for easier review.

ansible/roles/software/common/php-common/files/phpreload
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+service php{{ php_version }}-fpm reload
```
adborden: This won't get templated, since we're using the `copy` module. You'll want to move this to the `templates` dir and update the Ansible task.

ansible/roles/software/common/php-fixes/defaults/main.yml
```diff
@@ -4,7 +4,8 @@ php_cli_max_execution_time: 600
 php_cli_memory_limit: 2048M
 
 # Name of the PHP FPM service
-php_fpm_service_name: php{{ php_major_minor_version }}-fpm
+php_fpm_service_name: "php{{ php_major_minor_version }}-fpm"
 
 # Major/minor version of PHP
-php_major_minor_version: "7.0"
+php_major_minor_version: "{{ php_version }}"
+php_version: 7.3
```
adborden: FWIW, I like the `php_major_minor_version` name, because it's more descriptive. Technically, the PHP version would include a patch number (7.3.x) and I think the `php_version` was a poor choice on geerlingguy. And it's only now that I realize we could've reversed these variables to accommodate the third-party roles:

```
php_major_minor_version: "7.3"
php_version: "{{ php_major_minor_version }}"
```

Not a deal breaker though.

adborden (approved):
😎

.gitignore
```diff
@@ -62,4 +62,5 @@ gen-csr-out
 
 # ansible-galaxy vendored roles
 ansible/roles/vendor/*
-!ansible/roles/vendor/requirements.yml
\ No newline at end of file
+!ansible/roles/vendor/requirements.yml
+.vscode/settings.json
```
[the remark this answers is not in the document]
mogul: Handy! Added to my local config.

ansible/roles/software/common/php-common/files/phpreload
mogul: Good catch! Pushed a fix.

ansible/roles/software/common/php-fixes/defaults/main.yml
mogul: I thought about this too late while working on the branch... I agree, and made the change in another push.

adborden (approved):

adborden:
Strange, the build hung. I kicked off another one.

mogul:
Forgive me for merging my own PR, but I couldn't imagine anything in those last few commits changing @adborden's review.

PR #1067, opened 2019-10-25
Fix typo in cronjob name

mogul:
I noticed mail mentioning this job failing and this is the only place I was able to find it, so I'm hoping that fixing the typo here will help.


mogul:
Never mind, @adborden beat me to it in #1066

adborden:
Thanks for jumping in on this. I'll make sure to "call it" next time so we don't duplicate fixes :)

mogul:
Not your fault, it's mine for going to fix it as soon as I saw the mails, rather than finishing reading all my mail first and seeing your PR!

PR #1074, opened 2019-10-30
Update README.md

adborden:
Add a note about the tmux session.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,9 +59,14 @@ these requirements:
 
 Once you're SSH'd into the jumpbox, follow these steps for deploy.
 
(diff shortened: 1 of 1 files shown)


mogul (approved):

mogul:
Weird that the converge is failing, especially as it's not likely to be due to this PR!
```
    ckan-catalog-app-trusty    : ok=8    changed=4    unreachable=0    failed=1
```

Any ideas?

adborden:
Yeah, just seems like a network hiccup, the real failure is during the CKAN install, pip failed with an SNIMissing error on the SSL connection. I'm just going to retry.

Issue #1152, opened 2019-11-27
pod-dashboard S3 config is injectable


mogul:
## User Story

In order for pod-dashboard to make use of S3 in cloud.gov, pod-dashboard code must accept S3 credentials as injected configuration.

### Acceptance Criteria
- [x] GIVEN an S3 service is bound to the pod-dashboard app \
  AND the pod-dashboard app is configured to use S3 \
  WHEN code that depends on S3 access is exercised \
  THEN there are no errors. 

## Details

When running in the BSP, IAM roles control whether the host running pod-dashboard can write to the configured S3 bucket. This is anti-12-factor, because the code's behavior depends on a particular configuration of its environment. So: 

- The startup script should grab the S3 bucket key and secret from the cloud.gov environment and write it to .env on startup.
- The application code should make use of the configured key and secrets when accessing S3.

Other notes from Peter:
  * Credentials can be passed to the client by:
    * Using the environment variables , e.g.:
      * https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/getting-started_basic-usage.html
      * https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_environment.html
    * [Explicitly setting credentials](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_configuration.html#credentials), e.g.:
      ```php
      $credentials = new Aws\Credentials\Credentials('key', 'secret');

      $s3 = new Aws\S3\S3Client([
          'version'     => 'latest',
          'region'      => 'us-west-2',
          'credentials' => $credentials
      ]);
      ```

mogul:
I chose to go [the environment-variables route](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials_environment.html).

Still to do:
 - In `datagov-infrastructure-modules`
  - [ ] Add an S3 bucket for the CI environment.
 - In `datagov-deploy`
  - [ ] Add the S3 bucket credentials to the CI config and the `.env` file template

PR #1207, opened 2019-12-11
Dashboard webroot change

mogul:
This updates the Ansible deployment process to account for the changes in https://github.com/GSA/project-open-data-dashboard/pull/218 and https://github.com/GSA/project-open-data-dashboard/pull/219

This branch also 
* refactors the deployment to reduce the number of customized templates that need to be maintained
* adds configuration for the dashboard in the CI environment
* does a few minor cleanups.

This PR is a draft; it should not be merged until the two PRs above are merged.


ansible/dashboard-web.yml
```diff
@@ -25,6 +25,14 @@
   - { role: software/dashboard/dashboard-deploy, tags: [deploy, provision] }
 
 
+- name: Dashboard DB Migration
+  hosts: dashboard-web
+  serial: 1
+  become: no
+  roles:
+  - { role: software/dashboard/dashboard-db-migrations, tags: [provision, deploy, migrate] }
```
adborden: Will this break? I know we do some permissions changes in the clean-up step, so I would expect that migrations could fail to execute application code until the permissions are fixed... but maybe we fixed the permission issue on the app code with the `umask` change :man_shrugging:

ansible/inventories/ci/group_vars/dashboard-web/vars.yml
```diff
@@ -1,3 +1,27 @@
...
+default_host: dashboard.ci.datagov.us
+
+# MySQL credentials
+env_db_host: "{{ vault_env_db_host }}"
+env_db_user: "{{ vault_env_db_user }}"
+env_db_pass: "{{ vault_env_db_pass }}"
+env_db_name: "{{ vault_env_db_name }}"
+
+# AWS S3 bucket
+s3_bucket: bsp-ocsit-dev-east-appdata
```
adborden, line 13: Can you open an issue for this and add a comment? We need to create an S3 bucket as part of terraform and use that bucket here. This bucket won't work for the sandbox environments.

ansible/dashboard-web.yml
adborden: ... also, can we capture this in an issue: Dashboard should run the migrations on startup instead of it being part of deploy. I _think_ the startup script that @pburkholder added already does this in cloud.gov, but we want it to work the same here so that new migrations work consistently between environments.

ansible/roles/software/dashboard/dashboard-deploy/templates/.env
```diff
@@ -2,7 +2,17 @@ DB_HOST='{{ env_db_host }}'
 DB_NAME='{{ env_db_name }}'
 DB_PASSWORD='{{ env_db_pass }}'
 DB_USER='{{ env_db_user }}'
+# DB_DEBUG:
+#   TRUE: fail on database errors on migration, display to user
```
adborden, line 6: :sob: WTF code igniter!

ansible/roles/software/dashboard/dashboard-db-migrations/tasks/main.yml
```diff
@@ -2,12 +2,12 @@
...
 - name: Temporarily disabling keeping sessions in DB
   replace:
-    dest: "{{ current_source_symlink }}/application/config/{{ web_app_env }}/config.php"
+    dest: "{{ current_source_symlink }}/application/config/config.php"
     regexp: '(sess_use_database[^=]+)= TRUE;'
     replace: '\1= FALSE;'
 
 - name: Run migrations
-  command: php index.php migrate
+  command: php public/index.php migrate
```
adborden: Based on the comment in the `.env` template, shouldn't we set `DB_DEBUG=TRUE` so that the migration fails on error, and shows that error? We don't want the app to continue if the schema is in an inconsistent state.

ansible/dashboard-web.yml
mogul: I don't _think_ it will break. How can I tell other than running it in CI as I've been testing it?

And yes, I'll capture the "dashboard always runs its own migrations" in a separate issue.

ansible/inventories/ci/group_vars/dashboard-web/vars.yml
mogul: Sure, will do.

ansible/roles/software/dashboard/dashboard-deploy/templates/.env
mogul: Though it's a built-in config option for CodeIgniter's database class, we added the option to set it via the `.env`. It's set to default to FALSE, at least. 

Actually I just realized it's no longer used by the code, though... See [this line](https://github.com/GSA/project-open-data-dashboard/blob/ea376e75ac09770b4de979d4e05a7a75818fce2e/application/config/database.php#L92) carried in from the upstream CodeIgniter base. I like that default logic better; will yank this option out of `.env` entirely in another commit.

ansible/roles/software/dashboard/dashboard-db-migrations/tasks/main.yml
mogul: I believe the error will still hit the logs, so I don't think we _need_ to show it.  But yeah, we need this command to return a failure somehow. I'll try to add a CLI exit status when it fails.

ansible/dashboard-web.yml
adborden: If it were to break, it would probably break at the shell level running the command, with permission denied. If you've seen it run successfully, I think it's fine.

ansible/roles/software/dashboard/dashboard-db-migrations/tasks/main.yml
mogul: I looked into this more and [the existing call to `show_error()`](https://github.com/GSA/project-open-data-dashboard/blob/2e376fbf3f3e3494da774335e3be0114fe02ee52/application/controllers/Migrate.php#L29) (as well as [the implicit call made due to `migration_auto_latest`](https://github.com/bcit-ci/CodeIgniter/blob/0db8b60492cc98ef3c85b986fc11e9dac6d7ca6c/system/libraries/Migration.php#L181)) already [results in a call to `exit()` with an error status](https://github.com/bcit-ci/CodeIgniter/blob/0db8b60492cc98ef3c85b986fc11e9dac6d7ca6c/system/core/Common.php#L404-L417), so processing stops at that point.

ansible/dashboard-web.yml
mogul: > Dashboard should run the migrations on startup instead of it being part of deploy.

I've just made a change so [the dashboard does migrations on its own](https://github.com/GSA/project-open-data-dashboard/commit/2e376fbf3f3e3494da774335e3be0114fe02ee52), and [removed the migration task referenced here](https://github.com/GSA/datagov-deploy/commit/b9240af080fe24f7686a98f988d0487de34fa1aa). So no issue necessary!

ansible/inventories/ci/group_vars/dashboard-web/vars.yml
mogul: I'm going to do this in the course of resolving https://github.com/GSA/datagov-deploy/issues/1152, which I need to do before this gets merged in any case!

adborden (approved):

PR #1210, opened 2019-12-12
[catalog] fix tracking

adborden:
Removes current_env, which only works when --inventory is specified in
a specific way. Best to use an explicit variable instead.


adborden:
Closing as this isn't important right now.

Issue #996, opened 2019-09-28
Automated checks should catch errors in host file


adborden:
## User Story

As a developer, I want CI to catch errors in inventory variables and hosts so that when I make changes to these files, I'm confident that the changes are valid and correct.

## Details

Ansible 2.6.19 couldn't actually parse this https://github.com/GSA/datagov-deploy/pull/968/commits/2748ce59ca2aec7ff7824d6139f6417cc17bae1a

I'm hoping this can be supported with the standard ansible tools (e.g. `--syntax-check`) instead of rolling our own.

```
 [WARNING]:  * Failed to parse /home/gsa/projects/datagov/datagov-
deploy/ansible/inventories/production/hosts with yaml plugin: Syntax Error
while loading YAML.   did not find expected <document start>  The error appears
to have been in '/home/gsa/projects/datagov/datagov-
deploy/ansible/inventories/production/hosts': line 2, column 1, but may be
elsewhere in the file depending on the exact syntax problem.  The offending
line appears to be:  [catalog-harvester:children] catalog-harvester-v1 ^ here

 [WARNING]:  * Failed to parse /home/gsa/projects/datagov/datagov-
deploy/ansible/inventories/production/hosts with ini plugin:
/home/gsa/projects/datagov/datagov-
deploy/ansible/inventories/production/hosts:65: Expected group name, got:
inventory-web-v1 datagov_in_service=false
```

## Acceptance Criteria
- [ ] CI includes a test that checks inventory hosts file
- [ ] CI includes a test that checks inventory variables

adborden:
[HOST_PATTERN_MISMATCH](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#host-pattern-mismatch) was added in Ansible 2.8 which allows you to force a fatal error if there is an error matching the pattern to inventory. This could be used to solve part of the puzzle.

adborden:
[INVENTORY_ANY_UNPARSED_IS_FAILED](https://docs.ansible.com/ansible/latest/reference_appendices/config.html#inventory-any-unparsed-is-failed) added in 2.7 is also promising.

nickumia-reisys:
Wow... this level of server management is sooooo old!  Anyway, no more Ansible,
- https://github.com/GSA/data.gov/issues/3710

PR #1215, opened 2019-12-12
Include inventories in lint check

adborden:
Part of https://github.com/GSA/datagov-deploy/issues/996


adborden:
See https://github.com/GSA/datagov-deploy/pull/1889

PR #1216, opened 2019-12-12
Update CONTRIBUTING

adborden:


mogul:
I dig this change! Why is it still a draft?

mogul:
> I dig this change! Why is it still a draft?

adborden:
PERFECTION NEEDED.

Okay, fine we can get this merged. I wanted to say more about why UX is important and how we do that but.... yeah, it will always be a work in progress.

jbrown-xentity (approved):

PR #1218, opened 2019-12-12
Jenkinsfile

adborden:
Part of https://github.com/GSA/datagov-deploy/issues/1032

Parking the Jenkinsfile changes so far. Included are some **breaking changes**. @GSA/data-gov-team you will have to update your local development environment:
- Add a `.env` file in your datagov-deploy directory. It should contain:
  ```
  ANSIBLE_VAULT_PASSWORD_FILE=~/ansible-secret.txt
  ```
- Make target `update-vendor-force` is now just `vendor`.
- `requirements.yml` is now at the ansible directory and `ansible/roles/vendor` is completely git-ignored.

## How it works

I'm including the CircleCI steps, because this could be enabled in its incomplete state. Secrets need to be added and the CircleCI workflow should be un-commented. At a high level:
1. Commits trigger the usual CI jobs on CircleCI.
1. If this is a deploy branch (`develop` or `master`), after the CI checks pass, CircleCI tells Jenkins to run a deploy job.
1. Jenkins builds a docker image, `datagov/datagov-deploy`, locally.
1. Jenkins uses the docker image to run a sanity check.
1. Jenkins uses the docker image to run a playbook.

### CircleCI

CircleCI is still the de facto CI system.

We configure CircleCI to run the usual CI checks. On `develop`, if the CI checks pass, CircleCI will trigger the deploy job to deploy the platform to ci sandbox. The idea here is CircleCI is our de facto CI system and Jenkins is our deploy system which manages complexity of running ansible and deploying behind a firewall (BSP).

For the `master` branch (not implemented), CircleCI would trigger a separate job to deploy the platform into staging and then production.

And potentially we'd have release branches deploy to staging only, although the hope is with better automated CD, the release branches aren't necessary.

### Jenkins

Jenkins is our deploy system, hiding complexity of Ansible and deployments behind firewalls (BSP).

There will be two Jenkins instances, one in BSP MGMT VPC to deploy staging and production and one in our ci sandbox to deploy the ci environment.

Think of datagov-deploy as our platform deploy tool. When changes happen, we build a new artifact, a docker image `datagov/datagov-deploy`. We don't publish it to a registry (we could) because we only use it locally. The idea here is a deploy/release for datagov-deploy is building a new version of the deploy tool and using it to run the the platform playbook, `site.yml`. An application deploy/release is using the latest tested deploy tool (docker image) to run the application playbook.

[...]

PR #1220, opened 2019-12-12
Remove ansible.cfg

adborden:
This simplifies deploying with Jenkins so things like the Ansible Vault can be passed on the command line.


adborden:
This was rolled up into #1218

Issue #779, opened 2019-05-08
Configure MySQL app databases in sandbox


adborden:
## User Story
As a developer, I want MySQL application databases configured in sandbox environments from the ansible playbooks so that I can fully test the MySQL-based applications in sandbox.

## Acceptance Criteria
- [ ] Playbooks create the application databases in RDS using the [datagov-deploy-mysql](https://github.com/GSA/datagov-deploy-mysql) role

## Definition of Done
> Developers can provision machines in sandbox with terraform, the provision them with the ansible playbooks in order to have a working test environment.

nickumia-reisys:
No more FCS sandbox and ansible.
- https://github.com/GSA/data.gov/issues/3710

PR #1222, opened 2019-12-12
MySQL database provisioning in sandbox

adborden:
Part of https://github.com/GSA/datagov-deploy/issues/779


adborden:
Closing this. There are some issues with the postgresql role and I think we want to re-think the interface between platform responsibilities for provisioning application databases vs applications configuring their own databases.

Issue #1195, opened 2019-12-06
[Offboard] peter.burkholder@gsa.gov


adborden:
Peter was never fully onboarded, but we will go through the checklist to be thorough.

These tasks are performed by an Admin unless noted as an action for the outgoing team member.

- [x] Team member should move all Data.gov related documents to team drives
- [x] Team member should document any work in progress and schedule a hand-off for any remaining tasks
- [x] Remove team member from Data.gov GitHub teams
  - [data-gov-admins](https://github.com/orgs/GSA/teams/data-gov-admin/members)
  - [data-gov-support](https://github.com/orgs/GSA/teams/data-gov-support/members)
  - [data-gov-ckan-multi-partners](https://github.com/orgs/GSA/teams/data-gov-ckan-multi-partners)
- [x] Remove team member from AWS OPP account
- [x] Remove team member from [AWS sandbox access](https://github.com/GSA/datagov-infrastructure-live/tree/master/iam#new-users)
- [x] Remove team member from [jumpbox operators](https://github.com/GSA/datagov-deploy/blob/develop/ansible/group_vars/all/vault.yml)
- [x] Remove team member from Data.gov email lists
  - [Data.gov support list](https://groups.google.com/a/gsa.gov/forum/#!forum/datagov)
  - [Data.gov team list](https://groups.google.com/a/gsa.gov/forum/#!forum/datagovhelp)
- [x] Remove team member from [Data.gov calendar](https://calendar.google.com/calendar/r/settings/calendar/Z3NhLmdvdl9zcjZ0NG52YjRhOTNjNnNzdHRxYXAzbjZtMEBncm91cC5jYWxlbmRhci5nb29nbGUuY29t) and events
- [x] Remove team member from Uptrends via [email](https://docs.google.com/spreadsheets/d/1Z9Zpr1mpx-65i_fH2VTbVofPtidpLZs5cnkO0Jz53Vc/edit#gid=0)
- [x] Remove team member from [New Relic](https://newrelic.com)
- [x] Remove team member from [Docker Hub](https://cloud.docker.com/orgs/datagov/teams)
- [x] Remove team member from [Data.gov system accounts](https://github.com/GSA/datagov-account-management) following [Account Mangement Procedures](https://docs.google.com/document/d/1ipn-RX2Py1-jAAI_dbc6OdrK1j4Z8jKaya9C8RPyDa4/edit)
[...]

PR #1231, opened 2019-12-13
Remove peter.burkholder@gsa.gov from operators

adborden:
https://github.com/GSA/datagov-deploy/issues/1195


mogul (approved):
Inspected locally

PR #1251, opened 2019-12-26
[catalog] install from requirements-freeze

adborden:


jbrown-xentity (approved):
Yes, this is the correct way to implement pip install via ansible. Looks good.

Issue #581, opened 2019-02-11
Move test kitchen suites to molecule


adborden:
As part of the [Ubuntu bionic upgrade](https://github.com/GSA/datagov-deploy/issues/571), we're moving test kitchen suites to molecule where we can easily add bionic tests.

adborden:
@sstevens21 is working on the CRM, Dashboard, and WordPress roles.

adborden:
@sstevens21 is handing off the work. Latest commits will be pushed to https://github.com/sstevens21/datagov-deploy/tree/feature/move_apps_to_molecule

adborden:
Still waiting on the latest commits.

mogul:
Aaron's going to try to move this forward this sprint, particularly as the lack of automated testing for dashboard and Wordpress has been a hindrance.

mogul:
Aaron's planning to break these roles out into their own repositories to reduce build times/dependencies.

PR #1259, opened 2020-01-01
Remove test kitchen

adborden:

Remove test kitchen which isn't really used.


jbrown-xentity:
@adborden was Snyk using test kitchen for testing, is that why the check failed? I'm not completely comfortable with merging until that issue (either updating Snyk per documentation above or verification that system works) is resolved.

adborden:
Right, the Snyk check is failing because the Gemfile.lock was removed and you can safely ignore the check.

FYI, in general, this shouldn't hold up the a review. Any CI checks that should block a PR are set as "required" in GitHub. So you can safely approve the PR even if tests are failing. It's on the Author to resolve any required failing checks and get it merged.

jbrown-xentity (approved):
Looks clean

PR #1266, opened 2020-01-03
Snyk support for Pipfile

adborden:
Snyk currently does not support Pipfile through their web service. We have to
run the snyk CLI manually until it is supported in their service.


jbrown-xentity (approved):
Seems to work as expected, testing Snyk on circle ci for verification of every push.

PR #1271, opened 2020-01-03
Squash dashboard roles into a single role under test

adborden:
https://github.com/GSA/datagov-deploy/issues/581

This makes it simpler to write tests and refactor the dashboard deploy.

At this point we could move it into it's own repo https://github.com/GSA/datagov-deploy-dashboard.


jbrown-xentity (approved):
Looks good, though not tested locally

adborden:
Hmm, I left some of the permissive permissions (e.g. mode 0666) but I think the php-fpm runs as root, and the supervisor jobs run as ubuntu. So I think in general, we could restrict the permissions to ubuntu:www-data 0664 for the writable directories.

adborden:
I might revisit the permissions another time.

PR #1279, opened 2020-01-09
Add policy for PR reviews

adborden:
This came out from our retro today. Leaving this in draft status for folks to review and comment.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -271,7 +271,7 @@ the `hotfix/*` branch into `develop).
 ## Pull requests
 
 Developers should feel empowered to review each others code, even if you're not
-an expert on a particular application or feature. Any developers on the team can
+an expert on a particular application or feature. Any developer on the team can
 review any PR.
 
 What should you do when you review a PR?
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -268,4 +268,23 @@ Once the hotfix PR is merged, you should create a backmerge PR into develop (mer
 the `hotfix/*` branch into `develop).
 
 
+## Pull requests
+
+Developers should feel empowered to review each others code, even if you're not
+an expert on a particular application or feature. Any developers on the team can
+review any PR.
+
+What should you do when you review a PR?
+
+- Review the code for quality and consistency
+- Call out any breaking changes
+- Assert the **Definition of Done** is met
+  - Tests are written and passing in CI
+  - Documentation is written, if applicable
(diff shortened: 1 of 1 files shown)


adborden:
@jbrown-xentity you missed the discussion around this. Not sure if it captures everything you were wanting to capture.

jbrown-xentity (commented):
Want to note a number of exceptions:
1. There are a number of instances where CI is failing, but have been told it is ok to approve request
2. If the reviewer is not an expert on the purpose or language of the changes, how can they verify the code is in a deployable state? Should we approve if simple edits/fixes and seem like everything should work?

adborden:
Good points for clarification...

For CI, how about we drop the part about "passing tests in CI" since GitHub is enforcing that CI checks are passing. So adding something like:

> Any critical CI checks should be enforced by GitHub on protected branches (`develop` and `master`), so it's not required that CI checks are passing in order to approve a PR. Instead, it's important that tests have been added and they are running in CI.

On expertise:

> Data.gov encompasses many technologies (too many, in fact) and it's not practical to have everyone be an expert at everything nor to have only a single expert review all code in a specific domain e.g. PHP. Any developer should be able to follow the README in order to get a working system and should post on the PR if that doesn't seem to be the case.
>
> Ideally, our tests should give us good confidence that changes are working correctly. Even though that is currently not the case (we are working on building up our test coverage), it's not required to try out the code locally.

And finally, I want to note that this is why getting Jenkins setup is so critical. With continuous delivery to the sandbox on pushes to `develop`, that should give us good confidence as to whether or not `develop` is in a deployable state.

jbrown-xentity:
Absolutely, love those additions. And +1 to Jenkins on sandbox!

CONTRIBUTING.md
```diff
@@ -214,3 +214,22 @@ Branch | Deployed to | Frequency
 `develop` | AWS sandboxes | manual
 `release/*` | BSP dev | manual
 `master` | BSP prod | manual
+
+
+## Pull requests
+
+Developers should feel empowered to review each others code, even if you're not
+an expert on a particular application or feature. Any developers on the team can
```
mogul: s/developers/developer/

mogul (commented):
Looking forward to the additions from the conversation above, as that's extra-clear about what we're after!

adborden:
Okay, I've rebased and incorporated the changes discussed. This is ready for a final :+1:

mogul (approved):

PR #1281, opened 2020-01-10
[dashboard] use gsa.datagov-deploy-dashboard

adborden:
Use external GSA/datagov-deploy-dashboard role.

Part of https://github.com/GSA/datagov-deploy/issues/581

PR #1285, opened 2020-01-10
[wordpress] use external role

adborden:
This moves the wordpress role to GSA/datagov-deploy-wordpress.


adborden:
Going to re-create this with the #1281 to avoid merge conflicts.

PR #1286, opened 2020-01-11
[catalog] use recent pip and setuptools (release fix)

adborden:
Sometimes we see bugs on pip install due to old versions. Use a recent version
of pip/setuptools like we do in development.


jbrown-xentity (approved):
Would be good to get rid of all the warning flags for pip. Looks good, not tested.

PR #1288, opened 2020-01-15
Use galaxy roles for dashboard and wordpress

adborden:

This moves dashboard and wordpress roles to their own repositories, https://github.com/GSA/datagov-deploy-dashboard and https://github.com/GSA/datagov-deploy-wordpress.


jbrown-xentity (approved):
Much cleaner! Leaving to author to merge when validated in sandbox.

Issue #1255, opened 2019-12-30
Resolve XSS filter misconfiguration


mogul:
[details](https://docs.google.com/document/d/1_25KPRLpFC2r_2cJXoa91sUT_2llvf-qpbEX5-2pQog/edit#heading=h.w7m9co7bbwh4)

(migrated from https://github.com/GSA/datagov-incident-response/issues/56)

ydave-reisys:
<img width="1440" alt="Screen Shot 2020-01-24 at 3 01 50 PM" src="https://user-images.githubusercontent.com/3017376/73115331-8bff6600-3ef2-11ea-813a-2f9be3c885a3.png">

PR #1293, opened 2020-01-21
[wordpress] XSS fix

adborden:
https://github.com/GSA/datagov-deploy/issues/1255


mogul (approved):

mogul:
Approved code-wise, but it looks like this is getting hung up on the Snyk check.

PR #1316, opened 2020-01-29
Use upstream beats role

adborden:
Bumps gsa.datagov-deploy-common which uses the upstream beats role. This also
fixes the issue requiring us to use `--skip-tags filebeat` since the upstream
role handles this correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Once you're SSH'd into the jumpbox, follow these steps for deploy.
 1. Run the playbook from the ansible directory.
 
        $ cd ansible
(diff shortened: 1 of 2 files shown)


adborden:
I tested this on the staging jumpbox. Since the filebeat apt package was previously pinned with the previous role, I'm leaving it as a manual step to remove it. That means for the next deploy, as a one time step we should manually remove the package before running the playbooks.

    $ pipenv run ansible -m apt -a 'name=filebeat state=absent' all

adborden:
Whoops, I duplicated this in https://github.com/GSA/datagov-deploy/pull/1344.

PR #1317, opened 2020-01-29
[hotfix][catalog] increase number of CKAN workers (develop)

adborden:


jbrown-xentity (approved):
Backmerge, looks good.

Issue #1193, opened 2019-12-06
Expand dashboard to crawl all executive agencies under the OPEN Government Data Act


adborden:
## User Story

As a OMB, I want the Dashboard to cover all executive agencies under the OPEN Government Data Act so that any metrics the Dashboard provides can apply to all agencies under the law.

## Details
[meeting notes](https://docs.google.com/document/d/1v4nqG3kI1BM39wsN4E-v5Zr54KPZr8E945K7eF_QAvE/edit).

## Acceptance Criteria
- [x] Dashboard crawls list of agencies provided by OMB.
- [x] All agencies crawled are displayed on the front page of the Dashboard.

hkdctol:
Created a google doc, based on a list from OMB, of agencies beyond existing CFO Act agencies we should consider for this task. Will also indicate whether they already have a data.json harvest source. https://docs.google.com/spreadsheets/d/1fF_N4ZJqvGzm1G2bnrudnxq5Le9syVAUp_WwQ6l2KFk/edit?usp=sharing

PR #1321, opened 2020-01-30
Increase PHP memory to help crawls succeed

mogul:
Related to https://github.com/GSA/datagov-deploy/issues/1193

The expansion of the set of offices being crawled makes the crawl itself more memory-hungry and subject to being killed for using too much memory. The production hosts have more memory, and this change increases the headroom for the crawl to succeed.


jbrown-xentity (approved):
Looks good. Anyone reason for that specific number?

mogul:
Staging hosts have 2GB and production hosts have 4GB. We saw the crawls hitting the 2GB PHP max on the staging hosts, and figured giving them an extra GB may be enough for them to succeed without totally pegging the production hosts (which are also serving public traffic).

PR #1345, opened 2020-02-09
Pull hotfixes into develop

adborden:
Forgot to pull a few hotfixes into develop after merging them to master.


jbrown-xentity (approved):
Weird that there are 6 commits but only 2 line changes...

adborden:
Most of these are merge commits which are (usually) empty commits. The diff will only show content changes, so 2 lines makes sense if that's all that's changed.

PR #1368, opened 2020-02-12
Update onboard-team-member.md

adborden:
Add note that SSH key changes should be done as a hotfix so they can be applied to all environments immediately without having to wait for the next release cycle.


jbrown-xentity (approved):

jbrown-xentity:
Yeah, that's how we should have done it... hindsight is 20/20 and all that.

Issue #1183, opened 2019-12-05
Logrotate on inventory ckan logs is gzip'ing the archives


adborden:
Seeing log archives getting re-zipped e.g. `/var/log/inventory/ckan.error.log.2.gz.2.gz.1.gz.1.gz`

<img width="1074" alt="Screen Shot 2019-12-04 at 4 03 08 PM" src="https://user-images.githubusercontent.com/509703/70192239-f395f380-16af-11ea-8947-0a68f16d04dd.png">

Not sure if this affects other apps as well.

## How to reproduce

1. From inventory1p, `sudo ls /var/log/inventory`

## Expected behavior

Only the most recent logs, and archives are not re-zipped.

## Actual behavior

Many archives, each, re-zipped and re-zipped.

adborden:
Trying to deploy to staging and Inventory is complaining "No space left on disk". It has 7.8G available for /var/log, but I think this error could be referring to inodes or similar. I'm running this to clear the gzipped files.

    $ sudo find /var/log/inventory -name '*.log.*.gz.*.gz' -delete


I suspect this might also be why we're seeing high memory alerts on inventory randomly.

PR #1432, opened 2020-03-03
fix inventory logrotate

FuhuXia:
stop logrotate generating million of files by  gzipping files that has already been gzipped.
https://github.com/GSA/datagov-deploy/issues/1183


jbrown-xentity (approved):
Still cannnot merge, will support review of testing soon

Issue #682, opened 2019-03-29
inventory-web 500 OperationalError SSL SYSCALL error: EOF detected


adborden:
Sometimes we see 500s from Uptrends. Seems to be one of the hosts having a database issue:
```
[Fri Mar 29 00:21:20.728001 2019] [:error] [pid 27740:tid 140422794897152] [remote 10.176.16.106:29884] Error - <class 'sqlalchemy.exc.OperationalError'>: (O
perationalError) SSL SYSCALL error: EOF detected
[Fri Mar 29 00:21:20.728126 2019] [:error] [pid 27740:tid 140422794897152] [remote 10.176.16.106:29884]  'SELECT beaker_cache.data \\nFROM beaker_cache \\nWH
ERE beaker_cache.namespace = %(namespace_1)s' {'namespace_1': 'fc49500cb6d749879e85af8df10c22cd'}
```

I've seen this over the course of several hours. It's strange that the connection doesn't get re-established and recover. Rebooting or restarting apache should be a fine work-around.

FuhuXia:
This happen again today. Did a survey and there were 8k of occurrences of this error in the past 3 days.  Every time it happens user will see a 500 error page. So we need to resolve it to have a stable inventory web app.

I think it has something to do with the size of table beaker_cache. Query shows this table has 2.8 million records dated all the way back to December of 2016.  If we can have a slim beaker_cache table, this issue should be resolved. According to [Beaker docs](https://readthedocs.org/projects/beaker/downloads/pdf/v1.6.2/), it is developer's responsibility to delete old entries. 

```
2.4.1 Removing Expired/Old Sessions
Beaker does not automatically delete expired or old cookies on any of its back-ends. 
This task is left up to the developer based on how sessions are being used, 
and on what back-end.

The database backend records the last accessed time as a column in the database
so a script could be run to delete session rows in the database that haven’t been
used in a long time.
```

mogul:
We need to regularly run a query to prune entries older than a certain date. We should add an Ansible task that sets up a cronjob to do that on one of the hosts.

mogul:
@FuhuXia is going through testing his PR for this on sandbox, staging, and then prod.

adborden:
Is cleanup of beaker sessions supported in upstream CKAN? Seems like CKAN should be the one supporting this. A script/cron doesn't translate well to cloud.gov and ultimately this is the application's responsibility.

FuhuXia:
[...]

PR #1440, opened 2020-03-09
add beaker-cache-cleanup cron to inventory

FuhuXia:
fix issue: inventory-web 500 OperationalError SSL SYSCALL error: EOF detected
https://github.com/GSA/datagov-deploy/issues/682


ansible/inventories/staging/hosts
```diff
@@ -53,10 +53,12 @@ inventory-web-v1
 inventory-web-v2
 
 [inventory-web-v1]
-inventory[1:2]d.dev-ocsit.bsp.gsa.gov
+inventory1d.dev-ocsit.bsp.gsa.gov usage=with_cron
```
adborden: Let's use specifically named variables here, like `crons_enabled` like we do for dashboard and wordpress.

ansible/roles/software/ckan/inventory/tasks/additional-tasks.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: set usage type
+  set_fact:
+    usage_type: "{{ hostvars[inventory_hostname]['usage'] | default('') }}"
```
adborden: This isn't needed if we're just using a simple well-named variable.

ansible/roles/software/ckan/inventory/tasks/additional-tasks.yml
```diff
@@ -36,3 +40,38 @@
...
+        dest: /root/.pgpass
+        mode: "0600"
+        owner: root
+        group: root
+
+    - name: schedule beaker-cache-cleanup cron
+      cron:
+        name: "clean beaker cache in db"
+        cron_file: ckan
+        job: /usr/local/bin/beaker-cache-cleanup.sh
```
adborden: Instead of running the script directly here, we should run a supervisor job from cron. That way supervisor will handle things like logging, retries, and serialization of jobs.

ansible/roles/software/ckan/inventory/tasks/additional-tasks.yml
```diff
@@ -36,3 +40,38 @@
...
+        minute: "5"
+        hour: "11"
+        user: root
+        state: present
+  when: usage_type == 'with_cron'
```
adborden: Instead of using `when` this should be using `cron`'s `disabled` parameter. then there is no need for the extra "remove beaker cache cleanup cron job" task.

ansible/roles/software/ckan/inventory/templates/beaker-cache-cleanup.sh.j2
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# https://github.com/GSA/datagov-deploy/issues/682
+# delete stale session rows in beaker_cache table.
+
+age_in_days=7
+LOCKFILE=/tmp/beaker_cache_cleanup.sh.pid
+
+if [ -e ${LOCKFILE} ] && kill -0 `cat ${LOCKFILE}`; then
```
adborden: [flock](https://linux.die.net/man/1/flock) simplifies this logic a little bit. If you use flock, you can just wrap your critical commands in a subshell and you don't need to worry about deleting the lockfile as part of cleanup. flock will automatically remove the lock when the commands complete.

adborden (changes requested):

ansible/inventories/staging/hosts
adborden: ... Also, the pattern we've been using other places is to not specify variables in the inventory, but instead as host_vars https://github.com/GSA/datagov-deploy/blob/develop/ansible/inventories/production/host_vars/dashboardweb1p.prod-ocsit.bsp.gsa.gov/vars.yml

adborden (commented):
Looks good, thanks. Can you add a few tests to make sure these files get installed with correct permissions?

jbrown-xentity (approved):
Tests are implemented and working.

jbrown-xentity:
@adborden looks like you have to approve your changes were made before merge can happen.

ansible/roles/software/ckan/inventory/molecule/default/tests/test_default.py
```diff
@@ -37,3 +37,22 @@ def test_apache2(host):
 
     assert apache2.is_enabled
     assert apache2.is_running
+
+
+def test_beaker_cache_cleanup(host):
+    script = host.file('/usr/local/bin/beaker-cache-cleanup.sh')
+    supervisor_conf = host.file(
```
adborden, line 44: These should be separate tests to be consistent with how we write other tests, but :man_shrugging:

adborden (approved):

ansible/roles/software/ckan/inventory/molecule/default/tests/test_default.py
FuhuXia: ok, will separate the tests in another PR. let us get this one thru for now.

PR #1443, opened 2020-03-09
Remove unused Pipfiles

adborden:
These files are unused. If/when we split these roles out into their own repos,
we can add them.


jbrown-xentity (approved):
Yeah, why were those ever in there?

adborden:
Probably when creating the molecule boilerplate, these file were generated.

PR #1448, opened 2020-03-11
Dynamic Menu request uses https

jbrown-xentity:
Quick update to mirror development in using https instead of http for dynamic menu content


adborden (approved):
Probably just want to rebase to pick up the CI and Snyk fix without the extra commit.

adborden:
Are you sure this resolves the issue? I'm trying the same fix but running into https://github.com/GSA/datagov-deploy/issues/1467.

adborden:
Hmm, I think I might be confusing myself. It looks like `ckanext.geodatagov.dynamic_menu.url_default` is not used at all and `ckanext.geodatagov.dynamic_menu.url` has the correct default in [ckanext-datagov-theme](https://github.com/GSA/ckanext-datagovtheme/blob/f82b7c3cb991e133c72fee769042f6491250f113/ckanext/datagovtheme/helpers.py#L82), so there is no need to specify it.

jbrown-xentity:
Yes, the default is correct. I wonder if this change that was made doesn't actually fix anything, and the ability for the app to get the dynamic menu is intermittent. That's why the logic is built the way it is, to cache it locally.

adborden:
I reopened https://github.com/GSA/catalog-app/issues/76 and I have a work around I'm going to tack onto https://github.com/GSA/datagov-deploy/pull/1469

adborden:
Without this fix, I'm still getting the error, which is very very strange.
```
ubuntu@catalog-web1tf:~$ ls -l /var/tmp/ckan/dynamic_menu/menu.json
-rw-r--r-- 1 www-data www-data 43186 Mar 18 20:49 /var/tmp/ckan/dynamic_menu/menu.json
```

And then if I add this fix ... it works? I'm going to cherry pick this onto the release branch.

PR #1458, opened 2020-03-14
[catalog] create CKAN tmp/lock directory

adborden:
Not sure how this has slipped by, this directory is missing on sandbox.

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/molecule/default/tests/test_default.py
+++ b/ansible/roles/software/ckan/catalog/ckan-app/molecule/default/tests/test_default.py
@@ -6,6 +6,16 @@ testinfra_hosts = testinfra.utils.ansible_runner.AnsibleRunner(
     os.environ['MOLECULE_INVENTORY_FILE']).get_hosts('all')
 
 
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
What is this tmp folder used for? Should it be documented as a variable?

adborden:
I'm not really sure what this directory is used for. [Beaker](https://github.com/GSA/datagov-deploy/blob/0d5b27e9812707b865f23b7608d0f7be742ccd3a/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_ckan_production.ini.j2#L48) uses it and it's [hard-coded](https://github.com/GSA/ckanext-datagovtheme/blob/f82b7c3cb991e133c72fee769042f6491250f113/ckanext/datagovtheme/helpers.py#L17) in ckanext-datagovtheme.

I'm having doubts about this fix. The intention was to fix the issue with menu.json https://github.com/GSA/catalog-app/issues/76 and https://github.com/GSA/datagov-deploy/issues/1467 but the directory not existing doesn't seem to be the issue.

Issue #1473, opened 2020-03-18
Create a CKAN role based on ckan-app to handle Inventory and Catalog


adborden:
## User Story

As a Data.gov Developer, I want a CKAN Ansible role that can be used for Inventory and Catalog so that I don't need to maintain several separate and inconsistent roles and to identify tasks that should be moved to the applications.

## Acceptance Criteria
- [ ] A new GitHub repository exists for the Ansible role
- [ ] CI is configured for the new role
- [ ] datagov-deploy is updated to use the new role
- [ ] Issues are created for any tasks that should be moved to the responsibility of the applications (e.g. installing a production extension like ckanext-s3filestore)

## Details

Currently we have several roles for Catalog and several roles for Inventory. They are doing similar things, but differently. Even though they are running different versions of CKAN, the installation process is similar enough that we should be able to create a single role that can be configured for both Inventory and Catalog.

The new role should be similar to https://github.com/GSA/datagov-deploy-pycsw, which creates an OS user for ckan, improved deployment directory rotation, a user-specific python and virtualenv install, and less code to maintain.

nickumia-reisys:
Ansible no longer used and in charge of this,
- https://github.com/GSA/data.gov/issues/3710

Deployments to cloud.gov are managed as separate users per space (development, staging, prod) and these users are directly integrated with Github Actions for automated deployments.

Issue #719, opened 2019-04-08
Ubuntu bionic rollout to catalog-web host(s)


adborden:
## User Story
As an operator, I need new Ubuntu Bionic hosts provisioned in all of our environments so that I can continue receiving security updates automatically from Ubuntu.

Part of https://github.com/GSA/datagov-deploy/issues/571

Ubuntu 14.04 support ends [April 2019](https://www.ubuntu.com/about/release-cycle). We should move to Ubuntu 16.04 or 18.04. The latest 18.04 is preferred, but we should consider what kind of support exists for it.

## Details
Hosts are provisioned in these environments and play books are successfully applied.
- [x] bionic sandbox
- [ ] staging
- [ ] production

## Acceptance Criteria
- [ ] Bionic hosts are provisioned and operating correctly
- [ ] Bionic hosts reporting to our monitoring systems
- [ ] Old Trusty hosts are decommissioned and removed from Ansible inventory

adborden:
There's an issue with [M2Crypto](https://github.com/GSA/datagov-deploy/issues/758) a dependency for pysaml2/ckanext-saml2 that is blocking this upgrade.

adborden:
Getting this error in staging and sandbox:

From the host `curl -v -L -k localhost`. In `/var/log/ckan/ckan.error.log`:
```
[Sat Sep 28 16:19:41.522516 2019] [wsgi:error] [pid 19193:tid 139843254347520] [client 10.0.102.139:41506] Truncated or oversized response headers received from daemon process 'ckan': /etc/ckan/apache.wsgi
```

I bumped up [WSGIDaemonProcess](https://modwsgi.readthedocs.io/en/develop/configuration-directives/WSGIDaemonProcess.html#wsgidaemonprocess)'s header-buffer-size (to 32M) and response-buffer-size (to 64M) hoping to see more details about the response, but still get the same behavior.

adborden:
The above error is caused by https://github.com/GSA/datagov-deploy/issues/1375. The work around is to downgrade to pycsopg2 2.7.x.

Currently, I'm able to deploy catalog-web successfully https://catalog.bionic.datagov.us/

adborden:
I've also checked that pycsw is working https://catalog.bionic.datagov.us/csw

adborden:
Looks like we're not the only ones running into mod-wsgi problems https://github.com/ckan/ckan/issues/5430

mogul:
[...]

PR #1474, opened 2020-03-18
[catalog] roll saml2 and CKAN native login roles into ckan-app

adborden:
As part of https://github.com/GSA/datagov-deploy/issues/719, I'm running into some blockers to deploy CKAN on the bionic sandbox. This PR tackles some of those issues and should mean that we can deploy to sandbox successfully with the same playbook we use for BSP environments.

For a long time, we've been unable to run the same CKAN playbooks in BSP and sandbox due to the difference in authentication. BSP uses SAML2, while sandbox does not. The setup for SAML@/MAX.gov is painful and depends on an external team. In the past, we've used the ckan-native-login.yml playbook to make some updates to the CKAN install to remove saml2 and configure CKAN for native user/password login. That playbook hasn't worked recently and is a manual step.

This PR rolls up the saml2 and native-login roles into ckan-app itself. This reduces the number of duplicated variables and means that by running the catalog.yml playbook, you'll get the right configuration in sandbox and BSP environments.

Inventory still depends on the old saml2 role, so it was not removed in this PR. With this refactor, the next iteration would be to start combining Inventory and Catalog into a [new shared role](https://github.com/GSA/datagov-deploy/issues/1473).


jbrown-xentity (approved):

adborden:
Rebased to resolve conflicts.

PR #1485, opened 2020-03-20
Adds O and M issue template

mogul:


jbrown-xentity (approved):
Looks like all the things to me

PR #1496, opened 2020-03-25
Add paul.walker@datopian.com to sandbox

adborden:


jbrown-xentity:
@adborden didn't we decide we needed to make these changes as a hotfix to production?

adborden:
Yes, I was trying to cheat here. Since we deploy `develop` to sandbox,  and this is specific to sandbox, I was trying to skip having to do a hotfix. But thanks for being quick on this, I will submit a new PR based on master.

PR #1511, opened 2020-04-01
add who_ini_secret to ci harvester

FuhuXia:


adborden:
@FuhuXia do you know if you need this change in other environments as well (bionic, staging, production)?

adborden (approved):

FuhuXia:
@adborden Only needed in ci. I see other hosts already have this var defined.

PR #1533, opened 2020-04-08
rekey all vault.yml files with new ansible secret

FuhuXia:


mogul (approved):
Verified by running `find . -name vault.yml | grep -v '\./group_vars/all/vault.yml' | xargs  ansible-vault view` and then looking at each buffer presented to confirm they were properly decrypted.

PR #1550, opened 2020-04-16
stop tracking cron

FuhuXia:
no need to run the cron and consume resources if we know tracking is not working.


adborden (approved):

adborden:
Do you want to make this change as a hotfix, or attach it to the release so that it goes into production asap?

Issue #1611, opened 2020-04-29
Catalog rewrite making circular reference


jbrown-xentity:
Sandbox Catalog is making a circular reference redirect and cannot serve traffic. Due to [this line](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2#L60)

## How to reproduce

1. Go to https://catalog.ci.datagov.us/

## Expected behavior

Site serves /dataset page

## Actual behavior

Circular reference

PR #1612, opened 2020-04-30
[catalog] fix redirect loop on sandbox

adborden:
https://github.com/GSA/datagov-deploy/issues/1611

Allow a "default" configuration with no redirects so that standalone ckan is
still supported. Production and staging designate readwrite and readonly hosts.


adborden:
@jbrown-xentity this is ready for review.

jbrown-xentity (approved):
Looks good. Noting that we'll need to update test validation for wsgi with respect to gunicorn change coming, but it's correct as of now.

Issue #1536, opened 2020-04-09
Provision Redis (ElastiCache) for CKAN 2.8 in sandbox environment


adborden:
## User Story

As a Data.gov developer, I want Redis ([ElastiCache](https://aws.amazon.com/redis/)) provisioned in sandbox so that I can deploy Inventory CKAN 2.8 and meet its dependencies.

As a Data.gov developer, I want Redis ([ElastiCache](https://aws.amazon.com/redis/)) terraform module so that I can pass it over to BSP to demonstrate exactly what I need using infrastructure as code.

## Acceptance Criteria
- [x] GIVEN we run terraform in the CI environment
  THEN a Redis ElastiCache instance is created for CKAN 2.8
  AND an output describes the connection settings for use in the Ansible inventory
- [x] GIVEN I create a ticket for BSP to provision Redis ElastiCache
  AND I link to the terraform templates
  THEN BSP is able to provision Redis ElastiCache using our templates as a reference.

## Details

CKAN 2.8 depends on Redis so we'll need to provision a new one. We already have Ansible roles to provision a self-hosted Redis, but using the AWS service would be preferred so we don't have to manage it ourselves.

adborden:
https://gist.github.com/adborden/1f6ccddcf1b4f5f87d6fb1a737e142ef contains a gist with an example main.tf using the Redis module that we can hand off to BSP.

adborden:
@woodt would you QA this?

woodt:
So I think this looks fine ... two questions, mostly to cement my own understanding:

(1) We still need to add the catalog-next (really, catalog-harvester-next) redis host name to a group_vars?  I'd imagine we do that as part of one of the harvesting stories.

(2) Here's what the relevant portion of the ansible inventory looks like for sandbox, which seems OK to me but I want to confirm it is as intended:

```
 |--@redis:
  |  |--@redis-catalog:
  |  |  |--@tag_group_catalog_harvester:
  |  |  |  |--catalog-harvester-next1tf.internal.sandbox.datagov.us
  |  |  |  |--catalog-harvester1tf.internal.sandbox.datagov.us
  |--@redis-catalog-next:
  |--@redis-inventory-next:
```

adborden:
Hmm, that looks wrong and needs to be fixed in the datagov-infrastructure-* repos. I'll look into that. `tag_group_catalog_harvester` should no longer be part of `redis` or `redis-catalog`.

[...]

PR #1621, opened 2020-05-01
Move redis to its own group/playbook

adborden:
https://github.com/GSA/datagov-deploy/issues/1536

Decouple redis deployment from catalog. This allows us to configure multiple
redis hosts, separate from catalog (or inventory 2.8+) and it allows us to move
to a platform provided service (ElastiCache) by having no redis hosts in the
Ansible inventory.


mogul (commented):
I didn't see anything obviously amiss but it's also a pretty big change and hard to get my head around. What's the right way to see this working before I approve?

adborden:
FWIW, I tested this on sandbox, but besides that there's no great way to see this working.

It's mostly a refactor of where the role is called ([now redis.yml](https://github.com/GSA/datagov-deploy/pull/1621/files#diff-25a9082df3757df3f02bb844a7fdf7bdR5)) and where the variables live. Since the role itself hasn't changed (sitll in [GSA/datagov-deploy-redis](https://github.com/GSA/datagov-deploy-redis)), it's relatively low risk (and I would say a relatively minor change).

adborden:
... I don't mean to down-play it, we have a ton of bugs due to typos in inventory variables and playbooks. Unfortunately we don't have a great way to catch those except to apply them to an environment.

mogul (approved):

PR #1626, opened 2020-05-07
Snooze this whitelisted entry for another two months

mogul:
There's still no update on the same minor version with a fix, but we're still not vulnerable to this, so we may as well extend the exception for a while.


adborden:
I think this was addressed by the release https://github.com/GSA/datagov-deploy/pull/1633

Issue #1507, opened 2020-03-31
Provision Catalog sandbox environment for CKAN 2.8+


adborden:
## User Story

As a Data.gov developer, I want a sandbox environment provisioned side-by-side with the existing CKAN 2.3 environment so that I can deploy the new CKAN 2.8+ code to sandbox and ensure any iterative changes to the 2.8+ branch/repo can be deployed into a live environment without interrupting our ability to develop and maintain CKAN 2.3.

## Acceptance Criteria
- [x] WHEN I run terrafrom in the datagov-infrastructure-live repo
  THEN a new catalog environment based on Bionic along is provisioned
  AND the existing CKAN 2.3 environment is maintained.

## Details / tasks

We'll be deploying CKAN 2.8+ to the staging/production BSP environments. Having a sandbox environment allows us to continue to support CKAN 2.3, while developing CKAN 2.8+.

https://github.com/GSA/datagov-infrastructure-live is the top-level templates. A good place to start would be to copy `ci/catalog` into `ci/catalog_28` and then override variables.

https://github.com/GSA/datagov-infrastructure-modules contains the re-usable terraform modules. You _might_ need to edit these modules to allow to override variables.

The work here is similar to the Inventory issue https://github.com/GSA/datagov-deploy/issues/1481.

adborden:
@woodt let me know if you can connect to the hosts in sandbox (after adding SSH keys to datagov-deploy) and we'll call this QA'd :)

woodt:
Confirmed that the catalog web next instance is running.

woodt:
@adborden Trying to test connectivity to the catalog next RDS instance.  Should I be able to reach that host from the sandbox jump box?  E.g., I would expect:

```sh
nc -zv <vault_catalog_next_postgresql_login_host> 5432
```
to succeed, but it does not.  `vault_catalog_next_postgresql_login_host` came out of the datagov-deploy sandbox vault.

adborden:
No, not from the jumpbox, try that on the catalog-web or catalog-harvester hosts. The security groups are configured fairly strictly.

woodt:
I am able to ssh to the catalog-next-web1tf host.  From there, though, I can't seem to reach the catalog next RDS instance.  I pulled the host name out of the datagov-deploy sandbox vault and am testing with nc as above.  Getting no response at all.

@adborden

woodt:
[...]

PR #1627, opened 2020-05-07
[catalog] initial configuration for catalog-next

adborden:
https://github.com/GSA/datagov-deploy/issues/1507

This adds configuration for catalog-next generated from https://github.com/GSA/datagov-infrastructure-live/blob/master/sandbox/catalog-next.tf. Initiallly, these hosts are disabled in the catalog playbooks but still get common.yml plays applied.


woodt (approved):
This looks good to me.

Issue #1608, opened 2020-04-29
Potential DoS through high memory consumption


adborden:
_Please keep any sensitive details in [Google Drive](https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit#heading=h.4oadya3eps0d)._

**Date of report:** N/A
**Due date:** N/A

- [ ] Analysis has been performed and an issue has been linked to address other occurrences for this class of vulnerability\* (_link_)

_\* When a finding is identified, we create two issues. One to address the specific instance identified in the report. The other is to identify and address all other occurrences of this vulnerability within the application._

### Brief description
https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit#heading=h.4oadya3eps0d

PR #1630, opened 2020-05-08
remove datajson from catalog plugins

FuhuXia:
fix issue https://github.com/GSA/datagov-deploy/issues/1608


jbrown-xentity (approved):

adborden:
With datajson removed... how do we harvest from data.json harvest sources?

FuhuXia:
We still have datajson_harvest plugin.

PR #1645, opened 2020-05-12
Correct value for vault_catalog_next_postgresql_login_host

woodt:
Corrected the value to reflect the appropriate RDS instance name.


adborden (approved):
Reviewed vault locally.

Issue #1061, opened 2019-10-22
Inventory fails to serve requests in sandbox: S3 not configured


adborden:
S3 is not configured in sandbox environments, so requests fail to Inventory.

```
[Tue Oct 22 22:41:50.467031 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009] Traceback (most recent call last):
[Tue Oct 22 22:41:50.467046 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]   File "/etc/ckan/apache.wsgi", line 12, in <module>
[Tue Oct 22 22:41:50.467067 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]     _application = loadapp('config:%s' % config_filepath)
[Tue Oct 22 22:41:50.467075 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]   File "/usr/lib/ckan/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 247, in loadapp
[Tue Oct 22 22:41:50.467092 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]     return loadobj(APP, uri, name=name, **kw)
[Tue Oct 22 22:41:50.467098 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]   File "/usr/lib/ckan/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 272, in loadobj
[Tue Oct 22 22:41:50.467118 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]     return context.create()
[Tue Oct 22 22:41:50.467131 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]   File "/usr/lib/ckan/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 710, in create
[Tue Oct 22 22:41:50.467143 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]     return self.object_type.invoke(self)
[Tue Oct 22 22:41:50.467149 2019] [:error] [pid 18246:tid 140377939097344] [remote 10.0.102.24:29009]   File "/usr/lib/ckan/lib/python2.7/site-packages/paste/deploy/loadwsgi.py", line 146, in invoke
[...]

PR #1653, opened 2020-05-13
add s3 to sandbox inventory

FuhuXia:
fix missing s3 in sandbox environment https://github.com/GSA/datagov-deploy/issues/1061


adborden (approved):
:+1: is there a reason why the bucket name and prefix are secrets? Since we have private ACLs and permissions configured, I don't see a reason why we need to keep them in the vault. They are also incredibly guessable.

FuhuXia:
I saw mixed treatment of the bucket name and prefix, in some env it is plain string value but in staging and production it is vaulted value.  So to err on the safe side I put them in the vault.

PR #1668, opened 2020-05-15
naming the harvest-report cron

FuhuXia:
give it a name to avoid duplicate entries.


mogul (approved):

mogul:
Need to address the Snyk finding, but the code change is fine.

PR #1679, opened 2020-05-16
[hotfix] Disable CKAN 2.8 hosts

adborden:
Remove the CKAN 2.8 hosts from the catalog/inventory groups. We want to make
sure:
a) we aren't inheriting db settings that would conflict with older CKANs
b) want to make sure we're confident in sandbox before going forward


adborden:
Rebased to pick up snyk fix.

FuhuXia (approved):

Issue #1449, opened 2020-03-11
Proactive monitoring/restart of applications under pressure (monit)


adborden:
## User Story

In order to minimize the impact runaway processes on service availability, we want to automatically restart apache2 when threshold Y is observed.

## Acceptance Criteria

- [x] WHEN the monitored process on an application host is observed eating memory at a rate that will result in exhaustion in less than five minutes \
  THEN the monitored process is automatically restarted \
  AND the data.gov team is notified that the event happened \
  AND the notification includes a timestamp and a link to the relevant graph in New Relic

## Background

We sometimes see our catalog instances suddenly all climbing in memory (apache2/ckan) until eventually the host becomes unresponsive. This has been an ongoing issue for at least a year. 

As a workaround we've been restarting/reloading apache2 as described in the [runbook](https://github.com/GSA/datagov-deploy/wiki/catalog.data.gov#rapid-consumption-of-memory), but at this point it's reactive to an outage on our part, rather than proactively trying to avoid an outage, and SSH'ing in to do this is both a manual effort and potentially happens long after the problem appears.

## Sketch

- Try to make use of Monit to monitor and restart the relevant processes deployed on each host.

adborden:
@jbrown-xentity asked if we should commit the retry_ssh.sh script to version control. I'm putting it here as a gist, but not sure if it's worth adding to datagov-deploy.

https://gist.github.com/adborden/45b999e37f872aaeeab98724099ca6cf

adborden:
@FuhuXia is going to analyze the apache logs to determine if this is bot-related traffic. We'll start with a 4-hour timebox.

mogul:
Fuhu is running through hypotheses to try to narrow this down. Next step is to try to analyze the situation while it's actually happening.

mogul:
Blocked waiting for deployment of a fix to enable static asset caching, so that the logs are intelligible.

adborden:
Now that we've shipped a few fixes, moving out of blocked to Sprint backlog.

adborden:
Some tools available to us:
- [Upstart](http://upstart.ubuntu.com/wiki/Stanzas#limit) on Trusty allows us to set ulimits
- [Gunicorn](https://docs.gunicorn.org/en/latest/settings.html?highlight=memory#max-requests) allows a `max-requests` at which point it recycles the process. Catalog is not on gunicorn yet, and this isn't a fool-proof solution because it doesn't trigger based on memory
[...]

PR #1695, opened 2020-05-22
add script to monitor CPU/MEM usage

FuhuXia:


adborden (commented):
Can you link to the GH issue? I'm having trouble remembering why we wanted to add this. Don't we have this information in New Relic?

FuhuXia:
For this catalog memory issue https://github.com/GSA/datagov-deploy/issues/1449.
@adborden 
We can exam New Relic and get similar info **AFTER** an incident, but that won't help us to proactively restart catalog to prevent the incident. This script gives us the realtime, fine-grained readings based on which we build the threshold to trigger catalog restart.

adborden:
@FuhuXia I'm not seeing the restart in this script, though. Let's pick a value based on historic NR data and include the restart action in the script. I would add this to supervisor but maybe only enable on a single host if you want to test it first.

But let's get all the pieces implemented now with a best guess and then we can tune it later.

FuhuXia:
@adborden At this point the script only collects data. It has been manually deployed on catalog instances for a few days, waiting for next incident to happen. But I guess we can use historic NR data to roughly guess a threshold instead of waiting to build a perfect threshold. Let me calibrate the data collected from the script with NR data, and work on the threshold and restart script and supervisor conf.

FuhuXia:
@adborden The threshold is set as 99.50% of MEM usage for approximately 5 minutes. NR log retention is not long enough to be helpful. So this threshold is just best guess.

FuhuXia:
Sample output
Screenshot 2020-06-12 13 14 39

adborden:
How about we change the threshold to be 70% for 5 minutes? Just from my experience, I feel like that is more realistic.

FuhuXia:
Typical web instance tends to get stabilized at 90% after running for a few days. I see sometimes it peaks at 95% for more than 5 mins and it is still functioning and recovered itself back to 90%.

Screenshot 2020-06-12 16 18 11

ansible/roles/software/ckan/catalog/ckan-app/files/etc/supervisor/conf.d/supervisord_web_mem_cpu.conf
```diff
@@ -0,0 +1,7 @@
+; supervisorctl config file to restart apache when mem/cpu resource is critically low
+[program:cpu-mem-check]
+command=/usr/bin/cpu-mem-check.sh
+stdout_logfile=/var/log/ckan/cpu-mem-check.log
+redirect_stderr=true
+autostart={{ datagov_in_service | default(true) | ternary('true', 'false') }}
+autorestart={{ datagov_in_service | default(true) | ternary('true', 'false') }}
```
adborden, line 7: Let's create a new Ansible variable to enable/disable this. I think I'd like to run it on a single host to start and it would be nice to flip that on/off by Ansible configuration.

ansible/roles/software/ckan/catalog/ckan-app/files/etc/supervisor/conf.d/supervisord_web_mem_cpu.conf
```diff
@@ -0,0 +1,7 @@
+; supervisorctl config file to restart apache when mem/cpu resource is critically low
+[program:cpu-mem-check]
+command=/usr/bin/cpu-mem-check.sh
```
adborden, line 3: To be consistent, we install other scripts to `/usr/local/bin`

ansible/roles/software/ckan/catalog/ckan-app/molecule/catalog-next/molecule.yml
```diff
@@ -27,7 +27,6 @@ provisioner:
         # https://github.com/DavidWittman/ansible-redis/blob/21b0b6f9030275a2586baf591f322ce3348b2b2d/tasks/install.yml#L9
         redis_travis_ci: true
         ckan_catalog_next: true
-        ckan_uses_gunicorn: true
```
adborden, line 30: Can you explain what the issue is with the catalog-next test suite?

adborden:
Have you double checked the difference in your numbers with New Relic? I often see memory well under 70%. For example, right now they are looking all under 45% Memory used. That's why I think restarting services at 70% shouldn't be an issue.

Screenshot from 2020-06-12 14-17-33


I dug up an old screenshot from slack. https://gsa-tts.slack.com/archives/C4RGAM1Q8/p1580229915002800

Screenshot from 2020-01-28 08-46-35

The numbers are a bit rough, but when the memory issue hits, it hits fast. This shows it climbs from ~20% to unresponsive in less than 15 minutes. I think by the time memory hits 99.5% it's already too late.

ansible/roles/software/ckan/catalog/ckan-app/files/usr/bin/cpu-mem-check.sh
```diff
@@ -0,0 +1,53 @@
...
+  fi
+
+  # log
+  echo -e "$TIME\t$MEM%\t($MEM_TOP)\t$CPU_AVG%\t$CPU_CURRENT%\t($CPU_TOP)\t$STATUS"
+
+  # restart after max count of WARNINGs
+  if [ $warning_count_current  -eq $warning_count_max ]
+  then
+    echo "Restarting Apache"
+    sudo service apache2 restart
```
adborden, line 48: we're already running as root, so `sudo` shouldn't be necessary.

FuhuXia:
Ok, thanks for the NR info, good finding. Let me calculate MEM% using NR's way. I think NR include swap in the total.

adborden:
Hey @FuhuXia let's take a step back, my :spider: sense is tingling, I don't want to be tweaking this to handle all sorts of corner cases. There are so many existing tools in this space, why did we make the choice to go the custom script route?

What was your analysis of the existing options mentioned in https://github.com/GSA/datagov-deploy/issues/1449#issuecomment-625493772 and https://github.com/GSA/datagov-deploy/issues/1449#issuecomment-644240522?

FuhuXia:
Ok. Taking a step back. My initial intention for this script is monitor only. That is to identify a good indicator that tells the host is about to go down, and what process is responsible for it. I am currently trying out the Monit.

FuhuXia:
hold off for now.

PR #1697, opened 2020-05-22
solr-next hosts for CKAN 2.8

adborden:
https://github.com/GSA/datagov-ckan-multi/issues/296

Add hosts to Ansible inventory. This won't configure them with solr, but the
common playbooks will be applied to them.


adborden:
Testing on staging and will need a few tweaks...

PR #1719, opened 2020-05-29
Ansible updates for catalog-next harvester

woodt:
Adds support for enabling harvest workers for catalog-next sandbox.


ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -97,6 +97,14 @@ search.facets.responsible-party.title = Organizations
 ckan.search.solr_commit = true
 
 
+## Redis Settings
+
+# URL to your Redis instance, including the database to be used.
+# 'rediss' is not a typo!  It specifies redis proto over TLS
+
+ckan.redis.url = rediss://:{{ catalog_ckan_redis_password }}@{{ catalog_ckan_redis_host }}:6379/0
```
adborden: Let's make the redis url an Ansible variable. I think there might be some cases where we want to specify non-TLS redis and instead of adding a variable for that, just use a variable for the entire url.

adborden (approved):

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
woodt: Yes, nice!

woodt:
@adborden I added a redis URL ... I'm hoping this is OK.  Catalog Classic config does not use a URL for redis, so there is a little asymmetry here.  This has been tested on the sandbox .

PR #1724, opened 2020-06-01
Make sure the gunicorn (ckan) service is restarted.

woodt:


ansible/roles/software/ckan/catalog/ckan-app/tasks/web.yml
```diff
@@ -22,6 +22,12 @@
   service: name=supervisor state=started enabled=yes
   when: ckan_uses_gunicorn
 
+- name: Restart gunicorn / ckan service
+  supervisorctl:
+    name: catalog-web
+    state: restarted
+  when: ckan_uses_gunicorn
```
adborden: This should be done as an Ansible handler. Defined the [handler](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/inventory/handlers/main.yml#L6) and then attaching [notify statements](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/inventory/tasks/main.yml#L111) to the appropriate tasks (code was updated or config was updated).

adborden: I try to avoid using supervisor to restart an individual service because then we're sort of duplicating the definition of the service in two places (ansible tasks and the supervisor config). This helps separate app concerns from the platform concerns. Platform only cares that the new supervisor config is applied when something changes.

woodt: OK, I will use a handler and drive this off of notify's.

woodt: @adborden I think this is ready for another review.

adborden (approved):

PR #1725, opened 2020-06-02
Update plugins and ini file

avdata99:
- Updated for new extensions added
 - Updated INI file for DataGovTheme new settings


adborden (approved):

woodt:
This is great, thanks!  Only question I have is whether we really want the datastore and datapusher turned on.

adborden:
No, my understanding is that datastore/datapusher are not necessary for catalog as a metadata store. It would only be useful if we were storing datasets (and we are not). datastore/datapusher are not enabled on catalog.data.gov today (but are used for inventory.data.gov).

woodt (approved):
Approved.

avdata99:
Removed datastore and datapusher

PR #1758, opened 2020-06-16
Send notifications on all harvest jobs

avdata99:
Use the new config value defined for [GSA#315](https://github.com/GSA/datagov-ckan-multi/issues/315)

Is this the only place to add this setting for catalog next sandbox?


woodt (approved):
Ran the molecule converge test for the catalog-next scenario just to be sure.

PR #1770, opened 2020-06-18
Use updated dashboard role that omits SAML configuration

mogul:


mogul:
This should fix https://github.com/GSA/datagov-deploy/issues/1768

FuhuXia (approved):

PR #1776, opened 2020-06-19
Configurable parallelism

adborden:
Adds `datagov_serial` to control serialization where serialization matters. For
staging, we set this to 100%, meaning no serialization, because we don't mind if
services are unavailable during critical plays and would prefer faster deploys.

### DIFF
--- a/ansible/actions/reboot.yml
+++ b/ansible/actions/reboot.yml
@@ -1,6 +1,6 @@
 ---
 - hosts: '!jumpbox'
-  serial: 1
+  serial: "{{ datagov_serial | default(1) }}"
   gather_facts: false
   tasks:
     - name: Check if reboot is required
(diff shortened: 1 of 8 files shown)


FuhuXia (commented):
Staging is mentioned in the PR description but does not have datagov_serial set as 100%.

adborden:
Thanks! [This should've been staging](https://github.com/GSA/datagov-deploy/pull/1776/files#diff-02f8f9989444c8d462d65c62511d2ad6R76), not sandbox. I'll push a new commit

adborden:
I left the sandbox commit, there's only 1 host per ansible group, so `100% === 1` but doesn't hurt to leave it in.

FuhuXia (approved):

Issue #1777, opened 2020-06-19
Dashboard deploy fails on staging/production


adborden:
This is fixed in `develop` but `master` is not in a deployable state.

```
TASK [gsa.datagov-deploy-dashboard : Copying SAML config files] **************************************************************************************************************************************************************************************************************************************************************
failed: [dashboardweb1d.dev-ocsit.bsp.gsa.gov] (item=/home/ubuntu/datagov-deploy/ansible/roles/vendor/gsa.datagov-deploy-dashboard/tasks/../templates/saml/config/config.php) => {"changed": false, "checksum": "c1182447e3d5ad44183a5502f498d3b59eef5d2d", "item": "/home/ubuntu/datagov-deploy/ansible/roles/vendor/gsa.datagov-deploy-dashboard/tasks/../templates/saml/config/config.php", "msg": "Destination directory /var/www/dashboard/new/vendor/simplesamlphp/simplesamlphp/config does not exist"}
failed: [dashboardweb1d.dev-ocsit.bsp.gsa.gov] (item=/home/ubuntu/datagov-deploy/ansible/roles/vendor/gsa.datagov-deploy-dashboard/tasks/../templates/saml/config/authsources.php) => {"changed": false, "checksum": "c8bbea79cdb8abbec95842c823f9ee857105926e", "item": "/home/ubuntu/datagov-deploy/ansible/roles/vendor/gsa.datagov-deploy-dashboard/tasks/../templates/saml/config/authsources.php", "msg": "Destination directory /var/www/dashboard/new/vendor/simplesamlphp/simplesamlphp/config does not exist"}
        to retry, use: --limit @/home/ubuntu/datagov-deploy/ansible/site.retry
```

## How to reproduce

On the master branch, with the old version of datagov-deploy-dashboard (v1.0.3)
1. ansible-playbook dashboard-web.yml

## Expected behavior

Dashboard deploys successfully.

## Actual behavior

Error copying saml configs.

pjsharpe07:
Closing as this is being tracked in #1515

adborden:
I think this can be closed, it was fixed with the SAML hotfix https://github.com/GSA/datagov-deploy/pull/1778

PR #1778, opened 2020-06-19
[hotfix] Dashboard saml removal

adborden:
https://github.com/GSA/datagov-deploy/issues/1777

Mainly to pull in the bump to gsa.datagov-deploy-dashboard, but the other changes should be harmless as they:
- bump TLS in wordpress
- alter catalog-next configuration which is not deployed staging/production yet


mogul (approved):

adborden:
... and this is already on develop, so no backmerge necessary 😁

PR #1792, opened 2020-06-23
Replace harvest with datagov_harvest as catalog classic

avdata99:
Start using the `datagov_harvest` to override harvest source form creation
Now we are not able to create `waf-collection` harvester sources.
This is the [actual behaviour of catalog classic](https://github.com/GSA/datagov-deploy/blob/1cdb5dadecbc27a683d59e753dfeaab55bccc3ce/ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml#L16).
The form add this new fields:
image

### DIFF
--- a/ansible/inventories/sandbox/group_vars/all/vars.yml
+++ b/ansible/inventories/sandbox/group_vars/all/vars.yml
@@ -35,7 +35,7 @@ catalog_next_ckan_plugins_default:
   - image_view
   - text_view
   - recline_view
-  - harvest
+  - datagov_harvest
   - ckan_harvester
   - geodatagov
   - datajson


adborden:
Thanks @avdata99 can you link to where this extension is implemented? I want to pull in functionality from our custom extensions very deliberately. I think at some point after MVP, we'll want to circle back on all these various extensions for addressing tech debt. Therefore, I want to leave a trail of breadcrumbs so that we know where we want to look back to.... if that makes any sense.

adborden:
... or datagov_harvest is the new extension? :sweat_smile:

adborden (approved):

avdata99:
> Thanks @avdata99 can you link to where this extension is implemented? I want to pull in functionality from our custom extensions very deliberately. I think at some point after MVP, we'll want to circle back on all these various extensions for addressing tech debt. Therefore, I want to leave a trail of breadcrumbs so that we know where we want to look back to.... if that makes any sense.

Is defined at [geodatagov extension](https://github.com/GSA/ckanext-geodatagov/blob/master/setup.py#L30).


[This class](https://github.com/GSA/ckanext-geodatagov/blob/master/ckanext/geodatagov/plugins.py#L307) inherit and change behaviour of base harvesters.

We can't add waf-collection sources whitout the form defined at `package_form`.

Maybe we can define some issues to move this features in a better way into our new extension. We can then prioritize them and see want we need now and what could be posposed.
Is this a good idea?

avdata99:
> ... or datagov_harvest is the new extension?

I created a new issue to cover this: https://github.com/GSA/catalog.data.gov/issues/75

adborden:
This change was already made in development https://github.com/GSA/catalog.data.gov/pull/74

thejuliekramer:
@adborden @avdata99 This change does add the collection metadata url which we need for the waf collection harvester but this also causes the same error when trying to harvest a datajson source which worked previously, it changes other fields we need on the form like organization - we need to look into this more

thejuliekramer (changes requested):
We need to look into this more - it is breaking the datajson harvest flow

avdata99:
I try to add other types of sources and I get the same error defined at [Catalog#75](https://github.com/GSA/catalog.data.gov/issues/75).
We will need to fix 75 before merging this PR . I will close and add some reference to reopen after finish 75.

thejuliekramer (approved):

Issue #1796, opened 2020-06-24
Configure Postfix as an email sink in sandbox


adborden:
## User Story

In order to validate email features within sandbox without actually sending mail to users or having to maintain an SMTP service, data.gov developer wants Postfix to be configured as an email sink in the sandbox environment.

## Acceptance Criteria
[ACs should be clearly demoable/verifiable whenever possible. Try specifying them using [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).]
- [ ] GIVEN I run the common.yml playbook in staging
  WHEN I trigger an event that sends an email notification, e.g. a harvest job completes
  THEN I can inspect the postfix logs /var/log/mail.log on the host
  AND I can verify the message From, To, headers, and body
  AND the email is not actually sent

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
The sandbox is outside of our security boundary but we should ensure that if any other services are added, they do not expose additional ports or network connections and follow our security best practices.

## Sketch
There are several ways to do this in Postfix, and there are probably some non-Postfix tools that can help as well. The simplest solution that seems to work is to use [regexp canonical maps](http://www.postfix.org/ADDRESS_REWRITING_README.html) to rewrite the _envelope_ so that all mail is delivered to a local user (e.g. root) without rewriting the message headers. Mail can then be inspected using the `mail` command.

adborden:
The oefenweb.postfix role currently fails when specifying the `regexp` database type https://github.com/Oefenweb/ansible-postfix/issues/89

nickumia-reisys:
No more sandbox,
- https://github.com/GSA/data.gov/issues/3710

We use the `smtp-broker` for real email sending.  Our "test" for this is sending emails to datagovhelp@gsa.gov.  As a workaround for development use, we change all emails to be a fake email and don't actually care about the contents of the email.  If we did need to send a test email for content checking, we would use the generic datagovhelp email or use one of our gsa emails.
- https://github.com/GSA/data.gov/issues/4024
- https://github.com/GSA/data.gov/issues/3813

PR #1797, opened 2020-06-24
[postfix] update configuration for sandbox

adborden:
- Remove GSA relay
- Use the correct email From address
- Partially configure Postfix as an email sink


mogul:
@adborden James thinks this PR might not be needed anymore... Can you confirm and close if so?

adborden:
@jbrown-xentity can you elaborate? This removes the GSA SMTP relay which is not valid in sandbox.

pjsharpe07:
@adborden sorry, I misunderstood. This PR for when we pick up #1796

jbrown-xentity (approved):

Issue #1707, opened 2020-05-27
Catalog Tomcat upgrade


FuhuXia:
_Please keep any sensitive details in [Google Drive](https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit)._

**Date of report:** 05/01/2020 - nessus scan
**Due date:** 05/31/2020

### Brief description
Items [133845, 126125, 132413](https://docs.google.com/spreadsheets/d/1qInc2hrKBoBReklU9zbGvaeZcLY4-qmU0nae8nWZhEc) in nessus scan

adborden:
I removed tomcat from these hosts, which aren't supposed to have tomcat anyway.

```
ansible -m apt -a 'name=tomcat6 state=absent' catalog-web,inventory-web
```

adborden:
Also removed from these hosts since they are not currently in service.

```
ansible -m apt -a 'name=tomcat8 state=absent' catalog-harvester-v2
```

adborden:
I'm looking at the relevant CVEs and comparing versions to the Ubuntu repository and it looks like they are not available. Some investigation is necessary on what the right fix is, it's not as simple as apt-get upgrade.

FuhuXia:
@adborden Are we planning to move tomcat out of harvesters, into a dedicated bionic instance?

adborden:
Even if tomcat was running on bionic, it would not fix this.

FuhuXia:
Tried on local. Manually removed tomcat6 from Trusty catalog harvester instance and installed tomcat 8.5.56 on it. It is working.

mogul:
Is this one still in progress? (The compliance deadline for it is end of month five more days).

FuhuXia:
Yes, getting a PR ready. Hopefully can go to sandbox tomorrow.

FuhuXia:
Request send to BSP for fgdc2iso EC2 instances.
```
AMI: based on Ubuntu 18.04
instance type: t2.small (help to make better choice? )
name template: catalog-fgdc2iso1p
EBS:
 8GB root /
 8GB logs /var/log
The port 443 should be open internally to all catalog-harvester* instances in the same environment. 

Once the stack is created, please create one instance in DEV and one in PROD.

===========================
This instance will have tomcat 9 installed on it. Majority of  time it is sitting idle. When it is being used, it is accepting data from POST requests on port 443 from catalog harvester instances and return XML document.
[...]

PR #1810, opened 2020-06-26
add tomcat 9; refactor fgdc2iso to be an ansible role

FuhuXia:
Use latest tomcat 9 to avoid vulnerability found by nessus scan. https://github.com/GSA/datagov-deploy/issues/1707


ansible/roles/software/ckan/catalog/ckan-app/templates/etc_ckan_production.ini.j2
```diff
@@ -218,7 +218,7 @@ ckanext.spatial.common_map.subdomains = abc
 ckanext.spatial.common_map.attribution = Map tiles &amp; Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="https://creativecommons.org/licenses/by-sa/3.0/">CC BY SA</a>.
 ckanext.spatial.harvest.xslt_html_content = ckanext.datagovtheme:templates/xslt/iso-details.xslt
 ckanext.spatial.harvest.xslt_html_content_original = ckanext.datagovtheme:templates/xslt/esri-iso-fgdc-details.xslt
-ckanext.geodatagov.fgdc2iso_service = http://localhost:8080/fgdc2iso/
+ckanext.geodatagov.fgdc2iso_service = http://{{ catalog_ckan_fgdc2iso_host }}:8080/fgdc2iso/
```
FuhuXia: https required by bsp?

ansible/fgdc2iso.yml
```diff
@@ -0,0 +1,14 @@
...
+    - role: robertdebock.bootstrap
+    - role: robertdebock.core_dependencies
+    - role: robertdebock.java
+    - role: robertdebock.tomcat
+      vars:
+        tomcat_version9: "9.0.36"
+        tomcat_instances:
+          - name: "tomcat"
+            version: 9
+    - role: software/datagov-deploy-fgdc2iso
```
FuhuXia: temporary location. will be gsa.datagov-deploy-fgdc2iso

ansible/roles/software/datagov-deploy-fgdc2iso/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+fdgc2iso_tls_certificate_filepath: /etc/ssl/certs/datagov_host.crt
+fgdc2iso_tls_key_filepath: /etc/ssl/private/datagov_host.key
```
adborden: Since these don't really make sense as defaults and we're not actually managing these files, I would just make them required variables for the role, e.g. assert they are defined within the tasks at runtime and not include them in defaults.

ansible/roles/software/datagov-deploy-fgdc2iso/tasks/main.yml
```diff
@@ -4,21 +4,32 @@
     msg: saxon_license is required but has not been defined.
   when: saxon_license is undefined
 
-- name: include distribution-specific variables
-  include_vars: "{{ ansible_distribution_release }}.yml"
-
-- name: prepare packages needed for harvester
-  apt: name="{{ catalog_harvester_fgdc2iso_apt_packages }}" state=present
```
adborden: In terms of removing the existing tomcat package, what are you thinking? Should removing the package be a one-off as part of the next release?

ansible/roles/software/datagov-deploy-fgdc2iso/tasks/main.yml
```diff
@@ -4,21 +4,32 @@
...
+    - ROOT
 
+- name: install cron
+  apt: name=cron state=present
+
+  # TODO use cron module
 - name: copy clean-tmp-files cron
   copy: src=etc/cron.d/clean-tmp-files dest=/etc/cron.d/clean-tmp-files mode=0644
 
+- name: Copying nginix configuration
```
adborden:
```suggestion
- name: Copying nginx configuration
```

ansible/roles/software/datagov-deploy-fgdc2iso/templates/fgdc2iso.nginx.conf.j2
```diff
@@ -0,0 +1,19 @@
+server {
```
adborden: Let's use the [geerlingguy.nginx role](https://github.com/geerlingguy/ansible-role-nginx) instead. This will make it easier to share common nginx settings with dashboard and wordpress in the future. Currently dashboard and wordpress use geerlingguy.nginx for the package install only, but should use it for nginx configuration as well (obviously out of scope for this story).

There's a handful of missing things I see in this configuration:
- Prevent [header injection](https://github.com/GSA/datagov-deploy-wordpress/blob/f701837c70e3ff968470d5187521d31b5ef6fcd7/templates/datagov.nginx.conf#L9) in the redirect (not sure if nginx has since come up with a better, modern solution)
- Disable [TLS < 1.2](https://github.com/GSA/datagov-deploy-wordpress/blob/f701837c70e3ff968470d5187521d31b5ef6fcd7/templates/datagov.nginx.conf#L35)
- Disable weak ciphers

ansible/roles/software/datagov-deploy-fgdc2iso/templates/fgdc2iso.nginx.conf.j2
```diff
@@ -0,0 +1,19 @@
...
+
+server {
+  listen 443 default_server ssl;
+  server_name _;
+
+  ssl_certificate {{ fgdc2iso_tls_certificate_filepath }};
+  ssl_certificate_key {{ fgdc2iso_tls_key_filepath }};
+
+  location / {
+    proxy_pass http://127.0.0.1:8080/;
```
adborden: I'm not sure why, but I've always used [upstream](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#upstream) in combination with proxy_pass. Not sure if it pools connections with proxy_pass or if there are other benefits to using upstream :shrug:

ansible/site.yml
```diff
@@ -8,6 +8,7 @@
 ################
 - import_playbook: jumpbox.yml
 - import_playbook: solr.yml
+- import_playbook: fgdc2iso.yml
```
adborden: :+1: also should be added to the [Makefile](https://github.com/GSA/datagov-deploy/blob/28b155ca534076455967d92daa4567f5a31fb87d/Makefile#L6) for linting.

adborden (commented):
:+1: looking good so far @FuhuXia. It's already looking like a lot of work :) feel free to break it into additional stories if that makes sense.

I'm looking forward to seeing some molecule tests but the direction looks correct. My main callout is to move the nginx configuration into geerlingguy.nginx instead of managing an nginx.conf.

Also a tip: when you move this role to the new repo, molecule has an init command that will [scaffold the role for you](https://molecule.readthedocs.io/en/latest/getting-started.html#creating-a-new-role), just wanted to point it out in case you weren't aware.

ansible/roles/software/datagov-deploy-fgdc2iso/defaults/main.yml
FuhuXia: ok, but this means we need to repetitively define the same variable for all three environments?

ansible/roles/software/datagov-deploy-fgdc2iso/tasks/main.yml
FuhuXia: Yes, one-time ansible command to remove them after deployment.

ansible/roles/software/datagov-deploy-fgdc2iso/templates/fgdc2iso.nginx.conf.j2
FuhuXia: Cool. Did not know the role can do that.

FuhuXia:
@adborden I was aware of `ansible-galaxy` scaffolding, not `molecule`. Looks like molecule added file `.yamllint `and folder `molecule`.

ansible/roles/software/datagov-deploy-fgdc2iso/defaults/main.yml
adborden: Or you could define them once as a [playbook group variable](https://github.com/GSA/datagov-deploy/blob/develop/ansible/group_vars/all/vars.yml).

ansible/roles/software/ckan/catalog/ckan-app/templates/etc_ckan_production.ini.j2
adborden: Not by BSP, but by our SSP. All connections should be encrypted.

ansible/roles/software/datagov-deploy-fgdc2iso/defaults/main.yml
FuhuXia: Oh yeah. I thought there got to be a better way than declaring for all envs.

ansible/fgdc2iso.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Deploy Tomcat
+  hosts: fgdc2iso
+  roles:
+    - role: robertdebock.bootstrap
+    - role: robertdebock.core_dependencies
+    - role: robertdebock.java
+    - role: robertdebock.tomcat
```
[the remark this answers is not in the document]
FuhuXia: it is 0.0.0.0. Test added in molecule.

ansible/fgdc2iso.yml
```diff
@@ -43,3 +43,53 @@
...
+        # signed certs which should be valid.
+        # https://github.com/GSA/datagov-deploy/issues/900
+        validate_certs: false
+      retries: 3
+      delay: 10
+      register: result
+      until: not result.failed
+
+    - name: assert valid iso returned
+      assert:
```
adborden, line 80: Ah, I didn't know about `assert`, I've been using `fail` everywhere. Excellent test.

ansible/fgdc2iso.yml
```diff
@@ -43,3 +43,53 @@
...
+        that: "'<gco:CharacterString>ABC 123' in result.content"
```
adborden, line 81: [According to the docs](https://docs.ansible.com/ansible/latest/modules/assert_module.html), it looks like `that` should be a list. Since this is a string, it looks like you're iterating a string, which is always truthy, thus a false positive test assertion. Maybe instead:

```yaml
- name: assert valid iso returned
  assert: '<gco:CharacterString>ABC 123' in result.content

File: ansible/roles/software/datagov-deploy-fgdc2iso/molecule/default/tests/test_default.py [hunk 12]
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_tomcat(host):
+    tomcat = host.service('tomcat')
+
+    assert tomcat.is_running
+
+
+def test_port(host):
+    assert host.socket("tcp://0.0.0.0:8080").is_listening
```
adborden: tomcat should only listen to localhost, since nginx is handling the external connections.

ansible/roles/software/datagov-deploy-fgdc2iso/molecule/default/tests/test_default.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_cron(host):
+    cron = host.file('/etc/cron.d/clean-tmp-files')
+
+    assert cron.exists
+    assert cron.user == 'root'
+    assert cron.group == 'root'
+    assert cron.mode == 0o644
+    assert cron.contains('tmp*-*-*-*-*')
```
adborden: I think this should be a regular expression.
```suggestion
    assert cron.contains(r'"tmp\*-\*-\*-\*-\*"')
```

ansible/roles/software/datagov-deploy-fgdc2iso/molecule/default/tests/test_default.py
FuhuXia: the role comes with 0.0.0.0 and not configurable. I was relying on aws security groups to block 0.0.0.0:8080. If really needed, we can upload a custom server.xml file bind to ip 127.0.0.1 instead of default 0.0.0.0.

ansible/fgdc2iso.yml
FuhuXia: @adborden Nope, without `that` it is invalid syntax. When only one value, it is ok to use string as `that`'s value, same as the first example given in the [doc](https://docs.ansible.com/ansible/latest/modules/assert_module.html). 
`- assert: { that: "ansible_os_family != 'RedHat'" }`

ansible/roles/software/datagov-deploy-fgdc2iso/molecule/default/tests/test_default.py
FuhuXia: Not intended to be regex.  Should be exact string matching `tmp*-*-*-*-*`, I am changing it to the following to avoid confusion.
`assert 'tmp*-*-*-*-*' in cron.content_string`

ansible/roles/software/datagov-deploy-fgdc2iso/molecule/default/tests/test_default.py
FuhuXia: @adborden which approach to enforce 127.0.0.1 binding? Using lineinfile to change the server.xml in place, or providing a new server.xml to overwrite it altogether. Neither is my favorite.

FuhuXia: made a feature request to upstream, the new feature is implemented by the owner. now we can configure it to use 127.0.01.

FuhuXia:
Do not merge the PR yet. Pending on new instances setup on staging/production.

ansible/requirements.yml
```diff
@@ -50,6 +50,9 @@
 - name: gsa.datagov-deploy-common
   src: https://github.com/GSA/datagov-deploy-common
   version: v4.1.1
+- name: gsa.datagov-deploy-fgdc2iso
+  src: https://github.com/GSA/datagov-deploy-fgdc2iso
```
adborden, line 54: :+1: can you add me as admin?

adborden (approved):

ansible/requirements.yml
FuhuXia: added.

FuhuXia:
> Do not merge the PR yet. Pending on new instances setup on staging/production.

We are going to merge this so it can be deployed on Sandbox. If by the time of @pjsharpe07's next release, the new EC2's  are not set up in BSP yet, we can make a new PR with minor changes to keep existing tomcat intact.

Issue #1837, opened 2020-07-08
Remove jumpbox operators no longer in the operators list


adborden:
[datagov-deploy-jumpbox](https://github.com/GSA/datagov-deploy-jumpbox/blob/master/tasks/main.yml#L25-L36) will only remove/disable user accounts that have `active: false` in the list of operators. This is not very intuitive, a persons' first instinct would be to just delete the users from the list entirely.

## How to reproduce

1. Remove a user from `vault_datagov_operators_test`
1. Run the `jumpbox.yml` playbook

## Expected behavior

On the jumpbox:
`passwd --status $user` should show the account is disabled/removed.

## Actual behavior

On the jumpbox:
`passwd --status $user` shows the account is active

adborden:
This work isn't actually done, we can improve to make it less error prone.

#1838 is only about cleaning up old accounts, not addressing this issue.

jbrown-xentity:
[This thread](https://serverfault.com/questions/830302/ensure-only-specific-list-of-users-exist-with-ansible) provides an example implementation of exactly what we want to do. We could even implement the optional idea at the end on using SSH keys setup on each github account and get rid of our entire management of ansible keys... cc @adborden.

adborden:
> [This thread](https://serverfault.com/questions/830302/ensure-only-specific-list-of-users-exist-with-ansible) provides an example implementation of exactly what we want to do. We could even implement the optional idea at the end on using SSH keys setup on each github account and get rid of our entire management of ansible keys... cc @adborden.

Ah, minus the GitHub part\*, that looks like a good solution. I forgot we had created the operators group for our team members. I think you'll want to update our sudo config in the playbook, too, so that it's a single rule to allow the operators group sudo access. I think that just tightens up the disparate pieces.

\* With GitHub, I think we still have to manage the keys, we're just pulling them from GH instead of from our playbooks so I'm not sure that we gain anything. GH also allows multiple SSH keys, so if we want to restrict users to a single key we need a way to specify how to do that.

PR #1838, opened 2020-07-08
Disable old jumpbox operators

adborden:
These accounts were removed from the list in
af9d3d4ee53e7cf4d9fd92962a232fea754b5247, but the accounts were not actually
disabled/removed from the operating system due to
https://github.com/GSA/datagov-deploy/issues/1837


jbrown-xentity (approved):
Looks right to me. Are you sure you have the complete list of users now? How did you get this list @adborden?

adborden:
I was just scrolling through the commit log for any deletions (you do have to switch vault keys mid-way to the old key).

`pipenv run git log -p ansible/group_vars/all/vault.yml`

Issue #1560, opened 2020-04-17
Use a different ansible key for production


mogul:
## User Story

In order to minimize the blast radius from a compromised jumpbox or insider threat, we should use a separate Ansible key for the vaults for the production environment. 

## Acceptance Criteria
- [ ] GIVEN the common Ansible key IS installed locally
  AND the production Ansible key is NOT installed locally
  WHEN I run `ansible-vault view ansible/inventories/production/host_vars/*/vault.yml`
  THEN I see `ERROR! Decryption failed (no vault secrets were found that could decrypt) on` for each of the matching files.

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
[comment]: # "Our SSP says 'The Data.gov team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.' so please don't remove this section without care."
This change improves the security posture for data.gov as a whole since team members will not have access to production secrets as a baseline configuration... We can limit access to production secrets by only installing the production ansible-vault key on those hosts that actually need to access them (eg jumpboxes, for deployments).

PR #1853, opened 2020-07-10
[hotfix] create a separate key for sandbox environment

adborden:
https://github.com/GSA/datagov-deploy/issues/1560

This creates a new sandbox vault containing all the vault files in the sandbox inventory.

This will break developer workflows who are attempting to view/edit sandbox secrets. Developers should follow the recently updated [vault instructions](https://github.com/GSA/datagov-deploy/#ansible-vault) to get setup with the new sandbox key.


FuhuXia:
Verified vault is working after add new sandbox secret file.

jbrown-xentity (approved):
Confirmed access to various files using latest readme updates, all accessible.

adborden:
Updated the [vault rekeying instructions](https://github.com/GSA/datagov-deploy/wiki/Keypair-Rotation#ansible-vault).

adborden:
🤦 so I don't quite have a handle on this. Even though ansible/group_vars/all/vault.yml is encrypted with the production vault label, running the playbooks in sandbox still want to read it and fail. [AFAICT](https://docs.ansible.com/ansible/latest/user_guide/vault.html), you can't encrypt a single file with multiple vaults... so I think this means we have to get rid of ansible/group_vars/all/vault.yml. I think that's okay, but could use a pair to come up with an alternative.

jbrown-xentity:
This file was already a bit of an anti-pattern since we had to put in the key twice for the different environments, so maybe we just put them in the relevant inventories vault file? Would that require us to put them in both the staging and production vaults? Happy to :pear: sometime today or tomorrow...

adborden:
> This file was already a bit of an anti-pattern since we had to put in the key twice for the different environments, so maybe we just put them in the relevant inventories vault file? Would that require us to put them in both the staging and production vaults? Happy to  sometime today or tomorrow...

Right, if we did that, we'd actually have to add an operator key 4 times (production, staging, mgmt, sandbox). Trying to keep it down to two.

adborden:
Chatted with @jbrown-xentity and @FuhuXia last week on this. We should remove `ansible/group_vars/all/vault.yml` altogether.

adborden:
Replaced by https://github.com/GSA/datagov-deploy/pull/1928

PR #1854, opened 2020-07-10
ssl only for fgdc2iso; remove port 80

FuhuXia:
Don't see any reason leave port 80 open. Port 80->443 redirecting for fgdc2iso POST request is not working anyway.


jbrown-xentity (approved):

adborden:
I agree that we should be using https everywhere, but I don't see why we should disable port 80. All of our other services, public or internal should be configured the same.

Issue #700, opened 2019-04-03
Add yamllint to datagov-deploy


adborden:
As a developer, I want the inventory files checked in CI with linting so that I have confidence my inventory changes won't cause an issue when deployed.

nickumia-reisys:
Tangentially related on our new CI design,
- https://github.com/GSA/catalog.data.gov/pull/577

PR #1889, opened 2020-07-16
Include Ansible inventories in lint check

adborden:
I picked this up because I'm tweaking the hosts files for https://github.com/GSA/datagov-ckan-multi/issues/296.

I think this is the best we're going to get with https://github.com/GSA/datagov-deploy/issues/996.

If we provide the vault password to CI, we can do some more things like `ansible-inventory --graph` but I'm not sure it really buys us much more coverage. I think https://github.com/GSA/datagov-deploy/issues/700 would go pretty far.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,8 +38,10 @@ setup:
 	pipenv install --dev
 
 lint:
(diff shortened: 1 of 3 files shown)


jbrown-xentity (approved):
Looks good

Issue #1708, opened 2020-05-27
Truncated file download from Inventory using gunicorn


FuhuXia:
User reported xlsx file downloaded from Inventory couldn't be open. File appears to be corrupted. I tried a few times.  For 1-2 out of 10 times, the file downloaded appeared to be complete but the file size was slightly smaller than what it should be. When trying to open the file of smaller size, I got similar file corrupted error message. 

## How to reproduce

1. Go to [FY 2018 Federal Real Property Profile Data for Civilian Agencies](https://catalog.data.gov/dataset/fy-2018-federal-real-property-profile-data-for-civilian-agencies) on catalog.
2. Under 'Downloads & Resources', click 'Download' Button to download the xlsx file a few times.
3. Compare the file sizes, if any file is found smaller than others, try to open it.

## Expected behavior

File sizes are consistent after multiple download. 
## Actual behavior
Some download is smaller than others. Got error message when open it.

![Screenshot 2020-05-27 16 50 28](https://user-images.githubusercontent.com/1392461/83074324-f4488a80-a03f-11ea-88a3-13c5d210fa77.png)

FuhuXia:
This seems to be a newly surfaced issue. By its timing, maybe it is related to the Gunicorn we deployed recently? or maybe it is totally coincidence.

adborden:
We're getting inconsistent results when downloading data from the public internet. Running this command gives me a different and incorrect checksum each time. The expected MD5 checksum is `bfaa4e572e2ecea38ce3fa723b5bb58e`.

```
$ curl --silent 'https://inventory.data.gov/dataset/78b921cf-d865-44b6-84bc-9df85a5de642/resource/da3a910a-14f3-42b4-80d2-be3ce1360a5d/download/frpppublicdatasetfy1805232019.xlsx' | md5sum
```

```
0a4954bc558de0914c43390260fb9fc5  -
37a09a0922dc84d3d2cb0048ca5b9b3b  -
4cbeeaa62a30eff9c7e20e7fe4cc3194  -
```

From within our AWS environment in FAS Cloud Services, the command gives the correct checksum, both from the host as well as through Netscaler.

```
$ curl --silent -H 'Host: inventory.data.gov' http://localhost/dataset/78b921cf-d865-44b6-84bc-9df85a5de642/resource/da3a910a-14f3-42b4-80d2-be3ce1360a5d/download/frpppublicdatasetfy1805232019.xlsx | md5sum
[...]

PR #1923, opened 2020-07-24
Bring back gunicorn

FuhuXia:
Revert "gunicorn -> mod_wsgi" to bring Gunicorn back, also fix the truncated downloading issue #1708.


jbrown-xentity (approved):

ansible/roles/software/ckan/inventory/tasks/main.yml
```diff
@@ -100,15 +100,17 @@
...
-    - etc/ckan/apache.wsgi
     - etc/ckan/datapusher.wsgi
     - etc/ckan/datapusher_settings.py
     - etc/apache2/sites-enabled/ckan.conf
     - etc/apache2/sites-enabled/datapusher.conf
     - etc/supervisor/conf.d/gunicorn-web-app.conf
-  notify:
-    - restart apache2
-    - reload supervisor  # TODO the supervisor config files should be in their own task
+  notify: restart apache2
```
adborden, line 108: If `gunicorn-web-app.conf` is modified, don't you want to notify supervisor about the change?

ansible/roles/software/ckan/inventory/tasks/main.yml
```diff
@@ -100,15 +100,17 @@
...
+
+- name: load gunicorn
+  supervisorctl:
+    name: gunicorn-web-app
+    state: present
```
adborden, line 113: Personally, I prefer not to use the supervisorctl Ansible module because now the task name lives both in the configuration file and in the Ansible tasks. Is there an advantage to doing it this way?

PR #1928, opened 2020-07-27
[hotfix] create separate vaults for production and sandbox environments

adborden:
https://github.com/GSA/datagov-deploy/issues/1560

This creates a new sandbox vault containing all the vault files in the sandbox inventory.

This will break developer workflows who are attempting to view/edit sandbox secrets. Developers should follow the recently updated [vault instructions](https://github.com/GSA/datagov-deploy/#ansible-vault) to get setup with the new sandbox key.

I've updated the [documentation](https://github.com/GSA/datagov-deploy/wiki/Keypair-Rotation#ansible-vault) on the vault key rotation and have prepped the jumpboxes for the transition.


jbrown-xentity (approved):
Sanity check complete, all looks as expected with different environment keys and expected vault files...

PR #1945, opened 2020-07-29
[hotfix] Update New Relic collector

adborden:
Use the FedRAMP approved collector.

https://github.com/GSA/datagov-deploy/issues/1944

PR #1946, opened 2020-07-30
Add a timeout for harveting process

avdata99:
Related to [Multi#325](https://github.com/GSA/datagov-ckan-multi/issues/325#issuecomment-666335187)

This PR:
 - Add a timeout so harvest jobs are finished after some time.

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
@@ -197,7 +197,7 @@ set debug = False
 # is not specified here.
 
 # Mark as finished Jobs in 'Running' status after x minutes  
-ckan.harvest.timeout = 1440
+ckan.harvest.timeout = 1440  # 24 hours
 
 ckan.harvest.mq.type = redis
 ckanext.harvest.email = on
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
@@ -196,6 +196,9 @@ set debug = False
 # ckanext-harvest will use ckan.redis.url if redis configuration
 # is not specified here.
 
+# Mark as finished Jobs in 'Running' status after x minutes  
+ckan.harvest.timeout = 1440
+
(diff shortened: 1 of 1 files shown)


ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -196,6 +196,9 @@ set debug = False
 # ckanext-harvest will use ckan.redis.url if redis configuration
 # is not specified here.
 
+# Mark as finished Jobs in 'Running' status after x minutes
```
woodt, line 199: it would be great if the comment said "after 24 hours" so I don't have to do any math :-)

woodt:
So the CI tests are failing, but not for anything to do with this PR.

jbrown-xentity:
Same error as #1945, looks like this will block all merges until resolved. Not sure how it was introduced, and I cannot find the test. I believe it is related to recent changes for catalot-next and saml configuration... @adborden @starsinmypockets

jbrown-xentity:
Issue related to [this test](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/catalog/ckan-app/molecule/catalog-next/tests/test_default.py#L73), requirements were updated [here](https://github.com/GSA/catalog.data.gov/blob/master/ckan/requirements.txt#L77). Not sure if fix is to remove validation as we are doing changes to deployment, or if we need to revert change in catalog.data.gov requirements...

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -196,6 +196,9 @@ set debug = False
 # ckanext-harvest will use ckan.redis.url if redis configuration
 # is not specified here.
 
+# Mark as finished Jobs in 'Running' status after x minutes  
+ckan.harvest.timeout = 1440
```
adborden:
```suggestion
ckan.harvest.timeout = 1440  # 24 hours
```

adborden (approved):

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
adborden: 👍  I like this comment, we can add the 24 hour hint inline below.

PR #1947, opened 2020-07-30
add monit to manage catalog-web apache memory usage

FuhuXia:
To address high memory usage observed in catalog web instances https://github.com/GSA/datagov-deploy/issues/1449

### DIFF
--- a/ansible/catalog-web.yml
+++ b/ansible/catalog-web.yml
@@ -26,20 +26,30 @@
       vars:
         monit_cycle: "{{ monit_cycle_seconds | default(60) }}"
         monit_log_destination: /var/log/ckan/monit.log
+        monit_webinterface_enabled: false
         monit_mail_enabled: true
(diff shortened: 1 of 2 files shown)


adborden:
I was thinking we would add this to the common role so that we could use it
for any host, but this is a good start.

Is the last rule necessary? Why do we want to unmonitor after 5 restarts?

--

FuhuXia:
@adborden I put the last unmonitor rule as a precaution in case we set the rule wrong and get the server into infinite restarting loop. Also, if the Monit restarts the Apache2 relentlessly for 20 minutes and the server memory is still high, probably Monit is not going to be helpful any more. At the point manual inspection is needed. I came up with this rule on my own, it is not like I see some example done by other Monit users, so not really sure about it. Probably it will never be triggered anyway.

adborden (approved):

PR #1950, opened 2020-07-30
Capture attempts to restart harvester run

avdata99:
If starting harvester run fails, then show the error in logs `harvest-run: ERROR (already started)`

image


jbrown-xentity (approved):
Would be better to know than not know...
Will need to pick up testing fix to merge.

adborden:
I don't think we want this change. The harvester-run not starting because it's already starting is not an error condition, it's normal operating procedure. What we really want is to know when harvester-run has been running too long, and spamming the harvest-run log doesn't really help us.

avdata99:
Closed since is not what we need here

PR #1953, opened 2020-07-30
Hotfix for sandbox to avoid get stuck sending emails

avdata99:
Notify CKAN we are in a [test server](https://github.com/GSA/ckan/blob/datagov-newcatalog/ckan/lib/mailer.py#L57-L59) as [CKAN does](https://github.com/ckan/ckan/blob/master/test-core.ini#L90) as default for tests 

This should be only available for `etc_ckan_SANDBOX.ini.j2` file. Is not for production


ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -188,7 +188,7 @@ ckan.hide_activity_from_users = %(ckan.site_id)s
 #smtp.user = your_username@gmail.com
 #smtp.password = your_password
 smtp.mail_from = {{ catalog_ckan_email_from }}
-
+smtp.test_server = localhost:6675
```
adborden: Can you add a new Ansible variable `ckan_email_enabed: true` and then we can wrap this in an `if`
```jinja2
{% if ckan_email_enabled %}
smpt.test_server = true
{% endif %}
```

adborden (changes requested):

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
avdata99: Fixed

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -188,7 +188,9 @@ ckan.hide_activity_from_users = %(ckan.site_id)s
 #smtp.user = your_username@gmail.com
 #smtp.password = your_password
 smtp.mail_from = {{ catalog_ckan_email_from }}
-
+{% if ckan_email_enabled|default(False) %}
+smpt.test_server = true
```
woodt: Typo.  This should be `smtp`, not `smpt`.

woodt:
I think this should work, but it is a little confusing to me.  Maybe because I don't know what ckan does when smtp.test_server is set to true? :-)

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
avdata99:
```suggestion
smtp.test_server = true
```
typo

avdata99:
> I think this should work, but it is a little confusing to me. Maybe because I don't know what ckan does when smtp.test_server is set to true? :-)

[If 'smtp.test_server' is configured we assume we're running tests, and don't use the smtp.server, starttls, user, password etc. options](https://github.com/GSA/ckan/blob/datagov-newcatalog/ckan/lib/mailer.py#L56-L68)

ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -188,7 +188,9 @@ ckan.hide_activity_from_users = %(ckan.site_id)s
 #smtp.user = your_username@gmail.com
 #smtp.password = your_password
 smtp.mail_from = {{ catalog_ckan_email_from }}
-
+{% if ckan_email_enabled|default(False) %}
```
adborden: Sorry, I was not reading this closely when I made my suggestion.

1. If you set the variable in the [role defaults](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml), you don't need `default(False)`, which makes it less error prone if someone uses the variable elsewhere.
2. since this is enabling the test server and essentially _disabling_ email, let's call it `ckan_email_test_server_enabled` instead.
3. I wasn't reading the CKAN code closely, I thought it was just looking for a truthy value, but looks like it will actually try to open a connection? What's the right value to put in `smtp.test_server`?

avdata99: @adborden I added [these changes](https://github.com/GSA/datagov-deploy/pull/1953/commits/1473e5cb60fe72bdec99768e7bd3479a901a831a)
The test server will also [start a connection](https://github.com/GSA/ckan/blob/datagov-newcatalog/ckan/lib/mailer.py#L56-L68) but without TLS, user and password.

adborden: Hmm, so if CKAN still opens a connection and we have`smtp.test_server = localhost`, won't it still connect to the live postfix service on localhost:25?

adborden:
Replaced by https://github.com/GSA/datagov-deploy/pull/2013

PR #1965, opened 2020-07-31
[hotfix] new host certificates to add fgdc2iso hosts

FuhuXia:
Add fgdc2iso to host certificates so we can use port 443 for fgdc2iso service.


jbrown-xentity (approved):
Looks complete, will you test on staging before merging?

FuhuXia:
Ran playbook on staging, it passed host-level Ansible smoke test after a manual Nginx restart, but a fgdc source harvest job still failed with error `requests.exceptions.SSLError...CERTIFICATE_VERIFY_FAILED`. After more digging, it shows:

The following command gave same error.
```
(venv)root@catalog-harvester1tf:~# python -c "import requests;requests.get('https://catalog-fgdc2iso1tf.internal.sandbox.datagov.us/fgdc2iso/');"
```

This command worked:
```
(venv)root@catalog-harvester1tf:~# REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt python -c "import requests;requests.get('https://catalog-fgdc2iso1tf.internal.sandbox.datagov.us/fgdc2iso/');"
```

It turns out python's Request module is using its own ca_bundle located at `[venv]./lib/python2.7/site-packages/certifi/cacert.pem` rather than the system CA file, as people complained at [stakeoverflow](https://stackoverflow.com/a/56333265/13930770). Pip reinstalling module `certifi` did not help. Defining a shell variable REQUESTS_CA_BUNDLE with value `/etc/ssl/certs/ca-certificates.crt` in the shell as suggested in this [stackoverflow](https://stackoverflow.com/a/33717517/13930770) fixed it.

Not sure where to set this variable system-wide in Ubuntu. It looks like `/etc/profile` is a good place. Pushing another change.

adborden (approved):

ansible/roles/software/ckan/catalog/ckan-app/tasks/main.yml
```diff
@@ -14,6 +14,14 @@
       - swig
       - xmlsec1
 
+- name: set REQUESTS_CA_BUNDLE
+  blockinfile:
+    path: /etc/profile
+    create: true
+    block: |
+      # https://github.com/GSA/datagov-deploy/pull/1965#issuecomment-667562671
+      REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
```
adborden, line 25: Hmmm, I think my comment got lost. We might want to limit this to sandbox/staging since it is not needed in production. This is what we do in [pycsw](https://github.com/GSA/datagov-deploy-pycsw/blob/master/templates/supervisor_worker.conf.j2#L7-L10).

The reason is because the OS CA bundle can become "outdated" which is why `requests` bundles its own CAs via `certifi`... I'm afraid we might see harvester complain that certificates are untrusted.

adborden: ... And let's add a TODO comment to move this to the application environment (via .env or $HOME/profile), rather at the system level.

FuhuXia:
will merge into develop branch and go with regular release. closing this one.

ansible/roles/software/ckan/catalog/ckan-app/tasks/main.yml
FuhuXia: `REQUESTS_CA_BUNDLE` does not work as expected. It works when user runs python code on command line, but no by ckan harvester. We have to either find a way to append the GA certificate into `certifi/cacert.pem`  so Requests module can load it, or add the `REQUESTS_CA_BUNDLE` into ckan `virtualenv` environment variable. Issue #1999 is created to address this.

PR #1981, opened 2020-08-05
catalog-next use repoze.who 2.3

FuhuXia:
eliminate circleci error `AssertionError: assert '2.3' == '2.0'` on catalog-next.

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/tasks/main.yml
+++ b/ansible/roles/software/ckan/catalog/ckan-app/tasks/main.yml
@@ -98,7 +98,9 @@
     virtualenv: "{{ project_source_new_code_path }}"
     umask: '0022'
     state: present
-  when: not catalog_ckan_saml2_enabled
+  when:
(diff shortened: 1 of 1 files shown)


ansible/roles/software/ckan/catalog/ckan-app/tasks/main.yml
```diff
@@ -98,7 +98,9 @@
     virtualenv: "{{ project_source_new_code_path }}"
     umask: '0022'
     state: present
-  when: not catalog_ckan_saml2_enabled
+  when:
+    - not catalog_ckan_saml2_enabled
+    - "'v2' not in group_names"
```
adborden, line 103: Keying off of group names causes all kinds of issues, I think we really want to use an explicit variable to trigger this. It's also much easier to test in molecule when you're using an explicit variable.

However, since this is a quick fix to get the release out, we can go ahead with this.

adborden (approved):

PR #1997, opened 2020-08-06
Inventory-next sandbox integration

jbrown-xentity:
Not yet tested...


jbrown-xentity:
Not built in the right way, need a restart...

PR #2005, opened 2020-08-10
Bump Wordpress deploy

jbrown-xentity:
Bump wordpress


mogul:
@jbrown-xentity It looks like we may have to actually upgrade Ansible to get around the Snyk finding... What do you think?

mogul (approved):

PR #2014, opened 2020-08-11
Smoke Test retry bump

jbrown-xentity:


pjsharpe07 (approved):

ansible/inventory.yml
```diff
@@ -25,7 +25,7 @@
         follow_redirects: none
         status_code: 200
         validate_certs: false
-      retries: 3
+      retries: 10
```
adborden, line 28: This is the datapusher smoke test. You probably want to bump the inventory smoke test on Line 44 instead. That was the one you were having trouble with, right?

jbrown-xentity: Yep, created a [new PR](https://github.com/GSA/datagov-deploy/pull/2016)

PR #2017, opened 2020-08-11
ansible patch version bump to 2.8.14

FuhuXia:
Address Ansible uri module vulnerability found in https://github.com/GSA/datagov-deploy/issues/2009

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -17,11 +17,6 @@ ignore:
         reason: >-
           There is no fixed version for ansible. The vulnerable check mode is not used.
         expires: 2020-09-17T06:00:00.000Z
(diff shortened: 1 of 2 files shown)


FuhuXia:
hmmm... the commit which fixed the vulnerability is in 2.8.14 commit history, but Ansible does not announce version 2.8.14 is a fixed version for the vulnerability. We still get the snyk-test error. Let us hold it off for a few days and wait for the announcement, or wait for next 2.8 patch release.

jbrown-xentity (approved):
I've seen a delay between the reported fix and it propagating to Snyk for being marked as resolved, but it can't hurt to wait a day or 2...

Issue #1678, opened 2020-05-16
Enable builds for Inventory CKAN 2.8 on sandbox


adborden:
## User Story

In order to be confident that Inventory CKAN 2.8 will work in staging and production, data.gov developer wants deploys enabled in sandbox for Inventory CKAN 2.8.

## Acceptance Criteria
[ACs should be clearly demoable/verifiable whenever possible. Try specifying them using [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).]
- [x] WHEN I run the inventory.yml playbook in sandbox
  THEN the playbook runs successfully to completion

## Background
Inventory 2.8 is still [disabled on develop](https://github.com/GSA/datagov-deploy/blob/develop/ansible/inventory.yml#L3), so is not being deployed.

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
None

## Sketch
Need to model after catalog-next functionality:

- [x] Change inventory2_8 endpoints and variables to inventory-next
- [x] Apply ^^ to infrastructure in sandbox
- [x] Rebase inventory2_8 branch on current develop for datagov-deploy
- [x] Test deployment on sandbox manually via ansible

The following will be spun off into their own tickets to simplify the work being merged...
- [ ] Initialize the database
- [ ] Setup saml2 for enable/disable functionality for sandbox
- [ ] Fix datapusher configuration with [latest notes](https://docs.ckan.org/projects/datapusher/en/latest/deployment.html)
- [x] Fix gunicorn immediate crashing (SOLR configuration)
- [ ] Submit PR and finalize any issues with automated deployment via Jenkins

jbrown-xentity:
@adborden I can't remove the current inventory2.8 infrastructure, I don't have permissions... Can you either add permissions for infrastructure destruction to develop role, or pull the [in-work branch](https://github.com/GSA/datagov-infrastructure-live/tree/feature/inventory-next) and run `terraform destroy -target=module.inventory_2_8` as a one-off?

adborden:
@jbrown-xentity I'm worried that this is growing too much in scope. How about we limit the AC to just this:
> WHEN I run the inventory.yml playbook in sandbox
> THEN the playbook runs successfully to completion

and we set `datagov_in_service: false` so that even though it's not completely working, we're at least building and deploying. We can also add additional variables if we need to disable certain deploy tasks, e.g. datapusher.

[...]

PR #2020, opened 2020-08-12
Inventory-next for Sandbox

jbrown-xentity:
Manually pulls in some updates/tools from #2004, and refactors for new best practices.
To address #1678


jbrown-xentity:
Runs on sandbox for both inventory and inventory-next... 
image

jbrown-xentity:
This is currently failing because molecule doesn't pick up on variables declared in `group_vars/environment` variable declarations, like `group_vars/inventory/vars.yml`. This syntax is useful because we can declare variables that are specific to a stack but common across all environments. The reason we can't set them at the `group_vars/all` level is because of the "next" functionality, we need variables to have different values between `inventory` and `inventory-next`, but consistent across sandbox, staging, and production. We are currently breaking from [Ansible Best Practice on inventory architecture](https://docs.ansible.com/ansible/latest/user_guide/playbooks_best_practices.html#directory-layout) by using essentially both architectures. How do we want to handle this moving forward @adborden?

adborden:
@jbrown-xentity I'm not sure what you mean when you say we're not following the best practices for the Ansible directory structure. I think we _are_ following that? Keep in mind the [variable precedence rules](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable). The inventory group vars (inventories/sandbox/group_vars/catalog-web) override inventory all variables (inventories/sandbox/group_vars/all). Maybe you're getting confused with the playbook variables (ansible/group_vars/all)? You almost certainly don't want to define things in [ansible/group_vars/inventory](ansible/group_vars/inventory-next/vars.yml). We can pair on it and I'm sure that will clarify.

The way we handled this in catalog-next is that we completely ignore the existing catalog classic variables. Assume inventory-next (or inventory-next-web) is completely separate from inventory-web. Think of it as a completely independent app.

Ignore playbook variables for a moment. And just think about vars at the Ansible inventory level (ansible/inventories/staging/group_vars/). All the [catalog-next specific vars](https://github.com/GSA/datagov-deploy/blob/develop/ansible/inventories/staging/group_vars/catalog-next/vars.yml) are overridden. In that case, I used  `catalog_next_` to prefix variables so they could live in group_vars/all and you can see them side-by-side with catalog classic. You don't have to do that. You could move the inventory (classic) variables out of group_vars/all to group_vars/inventory and that way group_vars/inventory-next inherits nothing from inventory. From there, you can start pulling up variables that are common. Or not, it's kind of a minor optimization.

Now, back to playbook variables ([ansible/group_vars/catalog-next](https://github.com/GSA/datagov-deploy/blob/develop/ansible/group_vars/catalog-next/vars.yml)). In the catalog-next case, there are some variables that will definitely apply to catalog-next hosts across all environments, like `ckan_catalog_next: true` or the `app_repo: https://github.com/GSA/catalog.data.gov`. That's pretty rare and the one case to use playbook variables should be used. Even looking back at this, we're defining the [`catalog_ckan_app_version: master`](https://github.com/GSA/datagov-deploy/blob/develop/ansible/group_vars/catalog-next/vars.yml#L6) here, which is wrong. There are definitely cases where you would want to override that, but having it as a playbook variable makes that difficult.

Hope that's helpful.

ansible/inventories/sandbox/group_vars/inventory-next/vars.yml
```diff
@@ -0,0 +1,53 @@
...
+
+# Inventory Datapusher
+inventory_datastore_ro_user: datastore_ro
+inventory_datastore_user: datastore
+
+# project path
+project_source_rollback_path: /usr/lib/ckan-rollback
+project_source_new_code_path: /usr/lib/ckan-new
+current_source_symlink: /usr/lib/ckan
+project_index_path: /usr/lib/ckan
```
adborden: These I would expect to live in the role defaults without any modification.

ansible/inventories/sandbox/group_vars/inventory-next/vars.yml
```diff
@@ -0,0 +1,53 @@
...
+
+# db credentials
+inventory_ckan_db_host: "{{ inventory_next_postgresql_login_host }}"
+inventory_db_user: "{{ inventory_next_postgresql_login_user }}"
+inventory_db_pass: "{{ inventory_next_postgresql_login_password }}"
```
adborden, line 20: It doesn't really matter since we stopped using the datagov-deploy-postgresql role, but the idea here is that postgresql_login_user is the _admin_ user, and then a separate application user would get created (by admin) for use by CKAN.

The advantage being it's easier to rotate credentials for the application (create a new user with new creds without affecting the existing user).

FWIW, in catalog-next staging, I did make this distinction and there is a separate admin user and ckan user. For catalog-next sandbox, the creds are shared.

ansible/inventories/sandbox/group_vars/inventory-next/vars.yml
```diff
@@ -0,0 +1,53 @@
...
+# other
+inventory_ckan_bucket_name: "{{ inventory_next_ckan_s3_bucket_name }}"
+inventory_ckan_bucket_prefix: "{{ inventory_next_ckan_s3_bucket_prefix }}"
+ckan_site_domain: "{{ inventory_next_ckan_service_url }}"
+
+inventory_app_repo_branch: inventory_ckan_2.8
+
+# TODO is this for all databases (inventory, datastore, datapusher)? Don't think we need this
+# postgresql_enable_extensions:
+#   - postgis
```
adborden: Yeah, you can remove this since we're not using datagov-deploy-postgresql (although it would make db initialization much easier)

ansible/roles/software/ckan/inventory/defaults/main.yml
```diff
@@ -49,6 +49,7 @@ inventory_datastore_ro_user: ckan_datastore_ro
 inventory_datastore_user: ckan_datastore
 inventory_db_name: ckan
 inventory_db_user: ckan
+inventory_python_home: "{{ python_home }}"
```
adborden: Role defaults are true defaults, so we should really set a value here (and not refer to a potentially undefined variable). I _think_ `/usr` is the right value here. This value is confusing, longstanding technical debt 😩

ansible/roles/software/ckan/inventory/templates/etc/ckan/production.ini
```diff
@@ -96,6 +96,11 @@ solr_url = http://%(solr_server)s:{{ inventory_ckan_solr_port }}/solr/{{app_type
 
 #ckan.simple_search = 1
 
+## Redis Settings
+{% if inventory_ckan_redis_password is defined %}
+ckan.redis.url = redis://empty_username:{{ inventory_ckan_redis_password }}@{{ inventory_ckan_redis_host }}:6379/0
```
adborden: If it's an empty username, you can just leave it empty :) This also better matches the value for catalog-next.

```suggestion
ckan.redis.url = redis://:{{ inventory_ckan_redis_password }}@{{ inventory_ckan_redis_host }}:6379/0
```

adborden: You might want to pull the URL out into its own Ansible value. If we end up sharing redis between catalog-next and inventory next, you'll want a different database value (redis://host:6379/1). Are you planning to add a new redis host?

adborden (approved):
Good to merge, some minor comments/questions.

ansible/inventories/sandbox/group_vars/inventory-next/vars.yml
jbrown-xentity: This makes more sense for staging/production, as those machines will stay more static. As we can rapidly deploy sandbox environments, having to build out extra credentials manually will take more time and documentation than is probably valuable in this instance. Noting here though that we will do this for staging and production...

Issue #1795, opened 2020-06-24
catalog-harvester triggers fatal Postfix error in sandbox


adborden:
```
Jun 24 00:54:37 catalog-harvester-next1tf postfix/smtpd[7530]: fatal: in parameter smtpd_relay_restrictions or smtpd_recipient_restrictions, specify at least one working instance of: reject_unauth_destination, defer_unauth_destination, reject, defer, defer_if_permit or check_relay_domains
Jun 24 00:54:38 catalog-harvester-next1tf postfix/master[7488]: warning: process /usr/lib/postfix/sbin/smtpd pid 7530 exit status 1
Jun 24 00:54:38 catalog-harvester-next1tf postfix/master[7488]: warning: /usr/lib/postfix/sbin/smtpd: bad command startup -- throttling
```

## How to reproduce

1. Added an email to the `email_list` for the [State of Illinois](https://catalog-next.sandbox.datagov.us/organization/about/state-of-illinois).
1. Trigger a harvest for [illinois json](https://catalog-next.sandbox.datagov.us/harvest/illinois-json).
1. Verified the job completes successfully, with 0 added, updated, deleted
1. Inspect /var/log/harvest_run.log and  /var/log/mail.log

## Expected behavior

Postfix log shows message was received (not delivered, since sandbox isn't actually configured for sending).
Harvester shows message "Sent email to ..." message.

## Actual behavior

Postfix prints the fatal error, does not queue the harvester email.
Harvester hangs, waiting for Postfix.

adborden:
Need to resolve the harvest-run job hanging. The error doesn't seem specific to harvester, postfix just needs a configuration value set.

adborden:
Seeing this on staging, catalogharvester1d. Could this be a bionic thing? the same fix should work, but I'm luke warm on enabling email in staging without disabling sending to external accounts (non data.gov folks).

adborden:
Confirmed we're still seeing this on catalogharvester1p, so harvest report emails won't go out until this is fixed. I'm going to configure an email sink until we're ready for launch to avoid sending out catalog-next harvest reports to users.

adborden:
Confirmed this is resolved on catalogharvester1p.

PR #2028, opened 2020-08-13
[postfix] fix bad TLS key path

adborden:
Spotted this when debugging https://github.com/GSA/datagov-deploy/issues/1795

Looks like a bad refactor. TLS key exists in /etc/ssl/private. Use the playbook
variable across all environments.


jbrown-xentity (approved):
Looks great

PR #2033, opened 2020-08-14
[catalog-next] add fgdc2iso service to sandbox

adborden:
Adds an fgdc2iso service for catalog-next in sandbox. For staging/production,
we can share the existing service.

### DIFF
--- a/ansible/inventories/sandbox/group_vars/all/vars.yml
+++ b/ansible/inventories/sandbox/group_vars/all/vars.yml
@@ -26,6 +26,7 @@ catalog_db_user: "{{ catalog_ckan_db_user }}"
 
 # catalog-next
 catalog_next_ckan_db_name: catalog_db_next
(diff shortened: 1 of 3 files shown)


jbrown-xentity (approved):
Looks good. Ideally we wouldn't need 2 versions of this app that is the same across both environments, I'm assuming it's an access issue?

adborden:
Yeah, the security groups only allow catalog-harvester (classic) access. I kinda rushed through this to unblock ckan-multi team. We should be OK to use one instance in staging/production (and eventually sandbox). For sandbox, we would have to update [datagov-infrastructure-modules](https://github.com/GSA/datagov-infrastructure-modules/blob/master/modules/catalog/main.tf#L206) to pull fgdc2iso out of the catalog module into its own module. The way it works now, you always get a unique fgdc2iso service per catalog instance. Even if we updated the security groups to allow access from the entire VPC, there would still be two fgdc2iso services provisioned.

adborden:
Opened https://github.com/GSA/datagov-deploy/issues/2034 to follow up.

PR #2035, opened 2020-08-14
Update retry

jbrown-xentity:
So, according to [documentation](https://httpd.apache.org/docs/2.4/howto/reverse_proxy.html), it looks like we should be specific in proxying to localhost instead of `0.0.0.0`. Also, we noticed [timeout issues when the app was starting up](https://github.com/GSA/datagov-deploy/issues/2030), causing apache2 to wait for 60 seconds. This moves that down to 10 seconds, setting described [here](https://stackoverflow.com/questions/23709832/ap-proxy-connect-backend-disabling-worker-for-127-0-0-1).

@adborden thoughts on the specific timing, do we want 10, 30, 0?

### DIFF
--- a/ansible/roles/software/ckan/inventory/templates/etc/apache2/sites-enabled/ckan.conf
+++ b/ansible/roles/software/ckan/inventory/templates/etc/apache2/sites-enabled/ckan.conf
@@ -10,7 +10,7 @@ WSGISocketPrefix /var/run/wsgi
 
     #gunicorn configuration
     ProxyPreserveHost On
-    ProxyPass / http://127.0.0.1:5000/ retry=10
+    ProxyPass / http://127.0.0.1:5000/ retry=1
     ProxyPassReverse / http://127.0.0.1:5000/
 
 
--- a/ansible/roles/software/ckan/inventory/templates/etc/apache2/sites-enabled/ckan.conf
+++ b/ansible/roles/software/ckan/inventory/templates/etc/apache2/sites-enabled/ckan.conf
@@ -10,8 +10,8 @@ WSGISocketPrefix /var/run/wsgi
 
     #gunicorn configuration
     ProxyPreserveHost On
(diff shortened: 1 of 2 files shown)


adborden (approved):
👍  can you also look at the gunicorn configuration? Similarly, gunicorn should only be listening on localhost:5000, and not be accessible to other hosts (everything goes through apache).

ansible/roles/software/ckan/inventory/templates/etc/apache2/sites-enabled/ckan.conf
```diff
@@ -10,8 +10,8 @@ WSGISocketPrefix /var/run/wsgi
 
     #gunicorn configuration
     ProxyPreserveHost On
-    ProxyPass / http://0.0.0.0:5000/
-    ProxyPassReverse / http://0.0.0.0:5000/
+    ProxyPass / http://127.0.0.1:5000/ retry=10
```
adborden: Hmm, you could even make this as low as 1 second 🤷

jbrown-xentity:
Added, will merge pending test completion.

ansible/roles/software/ckan/inventory/templates/etc/supervisor/conf.d/gunicorn-web-app.conf
```diff
@@ -1,5 +1,5 @@
 [program:gunicorn-web-app]
-command={{ project_source_new_code_path }}/bin/gunicorn --worker-class gevent --paste /etc/ckan/production.ini -b localhost:5000
+command={{ project_source_new_code_path }}/bin/gunicorn --worker-class gevent --paste /etc/ckan/production.ini -b localhost:5000 --proxy-allow-from="127.0.0.1"
```
adborden, line 2: fwiw, I think 127.0.0.1 is the [default](https://docs.gunicorn.org/en/stable/settings.html#proxy-allow-ips)

Issue #1478, opened 2020-03-19
catalog tracking stopped being updated


FuhuXia:
Page hits and resource hits are not being recorded into tracking_raw table any more. As a result, view count for any dataset is not updated.

## How to reproduce
run query
`SELECT * FROM tracking_raw ORDER BY access_timestamp DESC LIMIT 1;`

## Expected behavior

access_timestamp should be current, and it should increase with every page hit.

## Actual behavior

stops at 2019-08-06 19:26:04.573118

adborden:
Tracking was [explicitly disabled](https://github.com/GSA/datagov-deploy/issues/488) because it was never completely working and generating errors (https://github.com/GSA/datagov-deploy/issues/205) The web instances that would be doing the tracking are "read-only", in that they are configured with a read-replica db. Therefore any writes to the tracking table would fail.

There was a partial implementation to redirect tracking URLs to admin-catalog, which _does_ have a read/write connection, but that implementation isn't working. Ultimately, the easiest thing to do was to disable tracking.

FuhuXia:
Fix is verified on prod.  Table `tracking_raw` is getting new entries now.

FuhuXia:
This has stopped working. The tracking only collects entries on admin site. Visits to read-only sites are not be recorded.

FuhuXia:
Admin site is not responding with expected CORS header to read-only sites' OPTIONS preflight requests. Fix is on the way.

FuhuXia:
Fixed. Visiting a dataset in catalog site, the dataset url shows up in the DB `tracking_raw` table.
```
ckan=> SELECT url, access_timestamp FROM tracking_raw ORDER BY access_timestamp desc limit 10;
                                              url                                               |      access_timestamp
------------------------------------------------------------------------------------------------+----------------------------
/organization/cfpb-gov                                                                          | 2021-09-24 14:51:43.760857
/organization/about/cfpb-gov                                                                    | 2021-09-24 14:51:40.146926
/group/older-adults-health-data                                                                 | 2021-09-24 14:51:38.134446
/organization/cfpb-gov                                                                          | 2021-09-24 14:51:33.817134
/dashboard                                                                                      | 2021-09-24 14:51:32.880848
/organization                                                                                   | 2021-09-24 14:51:32.176748
/dataset/supplemental-colleges                                                                  | 2021-09-24 14:51:28.12767
/dataset                                                                                        | 2021-09-24 14:51:27.753474
[...]

PR #2059, opened 2020-08-19
Fix tracking url

FuhuXia:
For issue https://github.com/GSA/datagov-deploy/issues/1478


jbrown-xentity (approved):
Looks like all the readable/writable variables are already there just waiting to be used...

Looks good!

jbrown-xentity:
I realized after a second go round that this doesn't have any tests... Could we validate that the final files are correct/as we expect in some way?

FuhuXia:
> I realized after a second go round that this doesn't have any tests... Could we validate that the final files are correct/as we expect in some way?

yeah, let me do that.

PR #2062, opened 2020-08-19
Add trendmicro policy to ungrouped_catalog group

pjsharpe07:


ansible/inventories/production/group_vars/ungrouped-catalog-next/vars.yml
```diff
@@ -0,0 +1,5 @@
+---
+# TrendMicro
+trendmicro_policy_id_web: "{{ vault_trendmicro_policy_id_web }}"
```
adborden: I think it's more like this. You want to set the policy_id to the web_id.

```suggestion
trendmicro_policy_id: "{{ trendmicro_policy_id_web }}"
```

Issue #1177, opened 2019-12-04
Specifying application/json for CKAN API fails on Inventory


adborden:
This looks like a bug in CKAN. When application/json is specified, it
returns 400 and the message:

> Bad Action API request data: Invalid request. Please use POST
> method for your request

Specifying application/x-www-form-urlencoded seems to keep it happy, even though it's
parsing json.

I've only seen this with POST methods.

## How to reproduce

1. Create a data file for package_create, `package_create.json`:
    ```
    {"name": "neh-peer-review-evaluators-1970-present", "notes": "test", "tag_string": ["test"], "title": "NEH peer-review evaluators, 1970-present", "owner_org": "national-endowment-for-the-humanities"}
    ```
1. curl -v -X POST -H 'Cookie: auth_tkt=1' -H "x-ckan-api-key: $CKAN_API_KEY" -H 'Content-Type: application/json; charset=utf-8' --data @package_create.json https://inventory-datagov.dev-ocsit.bsp.gsa.gov/api/action/package_create

## Expected behavior

200

## Actual behavior

400
> Bad Action API request data: Invalid request. Please use POST
> method for your request

adborden:
As a workaround, you should url encode your json data and specify `Content-Type: application/x-www-form-urlencoded`

```python
import json
import urllib.parse
import requests

encoded_data = urllib.parse.quote(json.dumps(data_dict))
response = requests.post(url, data=encoded_data, headers={'Content-Type': 'application/x-www-form-urlencoded'})
```

adborden:
Found this commit which also works around the issue for datapusher https://github.com/GSA/datagov-deploy/commit/1c3be95f88afc52a102f6d92df1e20f4678b425b

jbrown-xentity:
Not a bug we will fix. Closing.

Issue #2037, opened 2020-08-14
Inventory-next Datapusher configuration updated


jbrown-xentity:
## User Story

As a data.gov developer, I want inventory-next to deploy datapusher using the latest configuration so that upstream repositories and best practices are being used as much as possible.

## Acceptance Criteria

- [x] GIVEN I ssh onto inventory-next \
  WHEN I view the `supervisorctl status` output \
  THEN datapusher is running

## Background
Related to #1678. Blocks #1675

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
None

## Sketch
- Will need to possibly create separate role to deploy [using gunicorn](https://docs.ckan.org/projects/datapusher/en/latest/deployment.html#deployment-with-gunicorn) and add tests. 
- Should be able to implement [upstream version](https://github.com/ckan/datapusher); if not straightforward then create separate ticket and implement [GSA current version](https://github.com/GSA/datapusher)
- Only implement this latest version for inventory-next, may need an `inventory-next` variable set up to implement.

PR #2073, opened 2020-08-21
Datapusher for inventory-classic and inventory-next

jbrown-xentity:
This refactors how datapusher is deployed, moving into it's own tasks file for ticket #2037.
Since datapusher is configured differently for inventory-next (using gunicorn, using upstream, etc), I split them into completely different workflows.
Since some of the deployment verification was different for datapusher, I moved inventory-next/bionic testing into it's own molecule scenario (similar to catalog-next). Both systems passed locally.
Needed to pull in [Saml2 updates for inventory](https://github.com/GSA/datagov-deploy/pull/2081) to validate this was working as expected and to be able to test on sandbox. 
**All testing has passed, this is ready to merge.**


ansible/roles/software/ckan/inventory/defaults/main.yml
```diff
@@ -19,10 +18,13 @@ inventory_datapusher_error_log: "{{ inventory_log_dir }}/datapusher.error.log"
...
 
-datapusher_pkgs:
-  - name: datapusher
-    repo: https://github.com/GSA/datapusher
-    requirements: requirements-freeze.txt
+datapusher_build_pkg_name: datapusher
+datapusher_build_pkg_repo: https://github.com/GSA/datapusher
+datapusher_build_pkg_requirements: requirements-freeze.txt
+datapusher_gunicorn_error_log: "{{ inventory_log_dir }}/datapusher_gunicorn.log"
+datapusher_next_repo: https://github.com/ckan/datapusher.git
```
adborden, line 25: 👏

ansible/roles/software/ckan/inventory/defaults/main.yml
```diff
@@ -19,10 +18,13 @@ inventory_datapusher_error_log: "{{ inventory_log_dir }}/datapusher.error.log"
 inventory_datapusher_access_log: "{{ inventory_log_dir }}/datapusher.access.log"
 inventory_gunicorn_error_log: "{{ inventory_log_dir }}/gunicorn.log"
 
-datapusher_pkgs:
-  - name: datapusher
-    repo: https://github.com/GSA/datapusher
-    requirements: requirements-freeze.txt
+datapusher_build_pkg_name: datapusher
+datapusher_build_pkg_repo: https://github.com/GSA/datapusher
+datapusher_build_pkg_requirements: requirements-freeze.txt
```
adborden, line 23: Good variable names.

ansible/roles/software/ckan/inventory/molecule/default/molecule.yml
```diff
@@ -14,14 +14,6 @@ platforms:
     image: ubuntu:trusty
     groups:
       - trusty
-  - name: ckan-inventory-app-bionic
-    image: ubuntu:bionic
```
adborden: 🤔   I wouldn't have removed this but it's correct that this configuration will never be run on bionic. I'd prefer to test all roles on both trusty and bionic where possible. That way we don't have to think about is this role supported on trusty vs bionic (less thinking or knowledge about the code is better).

Are you seeing that this scenario no longer works on bionic?

ansible/roles/software/ckan/inventory/tasks/datapusher-classic.yml
```diff
@@ -0,0 +1,39 @@
...
+  file: path="{{ datapusher_virtual_env }}" state=absent
+
+- include_tasks: build-pkgs.yml
+  vars:
+    build_pkg_name: "{{ datapusher_build_pkg_name }}"
+    build_pkg_repo: "{{ datapusher_build_pkg_repo }}"
+    build_pkg_requirements: "{{ datapusher_build_pkg_requirements }}"
+    virtualenv: "{{ datapusher_virtual_env }}"
+
+- name: "Datapusher json content-type temp fix (content-type: application/json does not work now)"
```
adborden, line 14: FYI, once inventory-next ships, we'll be able to close https://github.com/GSA/datagov-deploy/issues/1177.
```suggestion
# TODO remove this code to resolve https://github.com/GSA/datagov-deploy/issues/1177
- name: "Datapusher json content-type temp fix (content-type: application/json does not work now)"
```

ansible/roles/software/ckan/inventory/tasks/datapusher-next.yml
```diff
@@ -0,0 +1,93 @@
...
+  template:
+    src: "{{ item }}"
+    dest: "/{{ item }}"
+    owner: root
+    group: www-data
+    mode: "0640"
+  with_items:
+    - etc/ckan/datapusher.wsgi
+    - etc/ckan/datapusher_settings.py
+    - etc/supervisor/conf.d/gunicorn-datapusher.conf
```
adborden, line 73: Fwiw, I would just name this "datapusher" since gunicorn is just an implementation detail.

ansible/roles/software/ckan/inventory/templates/etc/supervisor/conf.d/gunicorn-datapusher.conf
```diff
@@ -0,0 +1,10 @@
+[program:gunicorn-datapusher]
+command={{ datapusher_virtual_env }}/bin/gunicorn -b 127.0.0.1:6000 wsgi:app
```
adborden: Similar to inventory, I think we want to use an async worker type. The default is probably fine for now though.

adborden (approved):
Looks great!

FWIW, datapusher is really a completely separate application from inventory. And by that logic it could/should be it's own role, playbook, its own OS user, its own logs dir, etc. This PR will make it much easier to separate it out into it's own role when we're ready to pick up that work. When we start moving datapusher to cloud.gov, that would be a good time to do that since it would be deployed separately from inventory and moving to a separate role would keep parity between cloud.gov and BSP.

ansible/roles/software/ckan/inventory/molecule/default/molecule.yml
jbrown-xentity: I think I was trying to be too efficient on tests, and have different scenarios for classic and next. Instead, I'm going to slightly refactor to pull back in bionic and only have next specific tests on the inventory-next scenario...

PR #2090, opened 2020-08-28
[hotfix] correct urls admin

FuhuXia:
Following https://github.com/GSA/datagov-deploy/pull/2075
Use correct admin urls in staging and production. 
The variable name `urls_private` is misleading. Renamed to `urls_admin`.


jbrown-xentity (approved):
All looks good, vault file confirmed.

PR #2092, opened 2020-08-28
[hotfix] monit slack email (develop)

FuhuXia:
take `datagov_slack_devops_email` out of the vault file for all environments.


jbrown-xentity (approved):
Looks good, setting default to null and setting it for staging and production.

PR #2093, opened 2020-08-28
[hotfix] monit slack email

FuhuXia:
take `datagov_slack_devops_email` out of the vault file for all environments.


jbrown-xentity (approved):
Looks good, sets up default localhost and sets the variable in the necessary environments.

PR #2101, opened 2020-08-31
reverting before long-running functionality

pjsharpe07:


jbrown-xentity (approved):
Looks good, any reason this is still a draft?

pjsharpe07:
I was going to try and confirm if the `long-running` is what caused the host going down. I wanted to have this PR ready if that's the case so as not to slow down the release.

Still need to wait on the host to come back before confirming.

PR #2183, opened 2020-09-24
Revert Datapusher to WSGI configuration

jbrown-xentity:
Related to #2186 . 
Addresses a number of bugs, including virtual env setup, port usage, etc.


pjsharpe07 (approved):

ansible/roles/software/ckan/inventory/tasks/datapusher-next.yml
```diff
@@ -12,54 +12,28 @@
...
+    build_pkg_branch: "{{ datapusher_build_pkg_branch }}"
+    build_pkg_repo: "{{ datapusher_build_pkg_repo }}"
+    build_pkg_requirements: "{{ datapusher_next_requirements }}"
+    virtualenv: "{{ datapusher_virtual_env }}"
 
-- name: install requirements for datapusher
+- name: install psycopg2
   pip:
-    requirements: "{{ datapusher_virtual_env }}/src/{{ datapusher_next_requirements }}"
+    name: "psycopg2"
```
adborden, line 32: Why not install this from a requirements.txt?

jbrown-xentity: This was built to use upstream, which does not contain psycopg2 but is required for how we setup our virtual environment. We will need to investigate why it is necessary for our virtual environment, and clean up or submit a PR upstream. Either way, leaving this breadcrumb to note that we shouldn't support this long term.

PR #2192, opened 2020-09-25
[inventory] deploy new metadata form to sandbox

adborden:
Deploy the latest dcat-usmetadata extension to sandbox.


jbrown-xentity (approved):
So in the short term, if we want to see inventory changes on sandbox, we need to merge with this branch, correct @adborden ?

PR #2203, opened 2020-09-28
Add Solr "next" cores

jbrown-xentity:
For staging only at this time


pjsharpe07 (approved):

adborden:
These solr cores should be added to the new solr hosts, in the [solr-next](https://github.com/GSA/datagov-deploy/blob/develop/ansible/inventories/staging/group_vars/solr-next/vars.yml) group. There's not going to be enough capacity on the existing solr hosts, so we want to be using a dedicated group of hosts.

Issue #1491, opened 2020-03-23
catalog-bsp site url is indexed by search engine


FuhuXia:
catalog-bsp.data.gov is not supposed to be a public site. However, we started to see Google is crawling catalog site at this url. 

## How to reproduce

Go to https://www.google.com/search?q=site%3Acatalog-bsp.data.gov 

## Expected behavior

No results should be returned

## Actual behavior

About 42,600 results returned.

adborden:
@FuhuXia do you have ideas on the best way to fix? Since we don't actually want to serve anything at catalog-bsp.data.gov and essentially only use it as a CNAME for the BSP origin, I'm thinking we should just have CKAN restrict Host to catalog.data.gov via Apache config. That way CKAN will return 400s for any requests with `Host: catalog-bsp.data.gov`, but any valid requests (`Host: catalog.data.gov`) will respond 200 as usual.

Then there is a one-time cleanup to tell search engines to purge their indexes.

FuhuXia:
Yes, that will work. Still, using host http://localhost should be allowed so we don't have to spoof ip for local testing. So the change can happen on port 443 conf only, or we do allowing anything except `catalog-bsp.data.gov` instead of blocking anything except `catalog.data.gov`.

adborden:
I prefer whitelists over blacklists for this kind of thing. It's usually more secure. So only allow localhost and catalog.data.gov. Otherwise folks can get creative in using our app to serve arbitrary hosts which is problematic.

For port 443, I wouldn't tackle that yet. There's an outstanding issue to move CKAN instances to serve exclusively on HTTPS, but that involves some BSP configuration https://github.com/GSA/datagov-deploy/issues/570

FuhuXia:
Cool. Whitelisting `catalog.data.gov` sounds good to me as long as `localhost` is allowed.

adborden:
Just for reference, we already implemented this for [nginx hosts](https://github.com/GSA/datagov-deploy-dashboard/blob/5d1b8d9a982418e5f0771682d7b1301c3ee33aef/templates/dashboard.nginx.conf#L19) and define the [whitelist in a single variable](https://github.com/GSA/datagov-deploy/blob/b96f9bed12a0f8301cf7d0701764b256980ff9d2/ansible/inventories/production/group_vars/all/vars.yml#L48).

mogul:
A short-term fix would be to add a robots.txt that only appears for catalog-bsp.data.gov to tell search engines not to index.
[...]

PR #2259, opened 2020-10-05
Add next urls to be ServerAlias

FuhuXia:
Redo https://github.com/GSA/datagov-deploy/pull/2258. 
This time catalog-next hosts are added as ServerAlias.
for https://github.com/GSA/datagov-deploy/issues/1491

### DIFF
--- a/ansible/inventories/staging/group_vars/catalog-next/vars.yml
+++ b/ansible/inventories/staging/group_vars/catalog-next/vars.yml
@@ -12,8 +12,8 @@ catalog_ckan_readwrite_configuration: readonly
 catalog_ckan_redis_host: "{{ catalog_next_ckan_redis_host }}"
 catalog_ckan_redis_password: "{{ catalog_next_ckan_redis_password }}"
 catalog_ckan_redis_url: "{{ catalog_next_ckan_redis_url }}"
-catalog_ckan_service_url: "{{ catalog_next_ckan_service_url }}"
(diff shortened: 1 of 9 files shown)


ansible/inventories/staging/group_vars/catalog-next/vars.yml
```diff
@@ -39,5 +39,5 @@ saml2_site_url: https://admin-catalog-next-datagov.dev-ocsit.bsp.gsa.gov/
 # Token for working with the Google Analytics API
 token_dat: "{{ catalog_next_ckan_token_dat }}"
 
-url_readonly: "{{ catalog_next_ckan_service_url }}"
-url_writable: "{{ catalog_next_ckan_admin_service_url }}"
+url_readonly: https://{{ catalog_host_public_next }}
```
jbrown-xentity, line 42: Don't these need to contain the `"` to encode the variable properly?

jbrown-xentity (commented):
Is this tested on sandbox?

ansible/inventories/staging/group_vars/catalog-next/vars.yml
FuhuXia: Only if you start the value with {{}}. This value starts with `https://`, so it is fine.

FuhuXia:
> Is this tested on sandbox?

Not yet, but It is tested on staging, which is a more thorough test than sandbox, so I think it is good to go to sandbox.

jbrown-xentity (approved):

PR #2274, opened 2020-10-13
Use async (gevent) worker model for gunicorn. Also include user/group…

CodeShtuff:
… settings for permissions consistency

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
@@ -7,4 +7,3 @@ stdout_logfile={{ catalog_gunicorn_error_log }}
 autostart=true
 autorestart=true
 redirect_stderr=true
-startsecs=0
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
@@ -1,7 +1,10 @@
 [program:catalog-web]
-command={{ project_source_new_code_path }}/bin/gunicorn --paste /etc/ckan/production.ini -b localhost:5000
+command={{ project_source_new_code_path }}/bin/gunicorn --worker-class gevent --paste /etc/ckan/production.ini -b localhost:5000 --proxy-allow-from="127.0.0.1"
 directory={{ project_source_new_code_path }}
+user=www-data
+group=www-data
 stdout_logfile={{ catalog_gunicorn_error_log }}
 autostart=true
 autorestart=true
 redirect_stderr=true
+startsecs=0


ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
```diff
@@ -1,7 +1,10 @@
...
 directory={{ project_source_new_code_path }}
+user=www-data
+group=www-data
 stdout_logfile={{ catalog_gunicorn_error_log }}
 autostart=true
 autorestart=true
-redirect_stderr=true
\ No newline at end of file
+redirect_stderr=true
+startsecs=0
```
adborden: 0 is no start timeout... Don't we want to set _something_? Maybe 60 seconds?

CodeShtuff: Thanks for the note @adborden !
Good question. The setting was grabbed based on finding it having been added to the settings for gunicorn on inventory previously (similar ticket reasons it seems) which I referenced for these entries to catalog: https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/inventory/templates/etc/supervisor/conf.d/gunicorn-web-app.conf

Of course that does not necessarily mean that the same settings should apply for catalog. 

In looking at the docs for the startsecs setting (http://supervisord.org/configuration.html), if I am interpreting it correctly, it is not so much a start timeout but rather a minimum time value that it checks against for it to be considered as having successfully started/run. I am not familiar enough with the normal processing of gunicorn as needed in catalog to understand what 'normal' operation looks like and if we want to denote no minimum runtime via the startsecs setting, or if we have a value in mind that should be used. Appears the default is "1" so we could also simply not define a different value.

Shall we go with not setting, to therefore use the default, and then we can adjust if we find that new issues are discovered?

adborden: I see. So the default is 1. @FuhuXia do you know why we want to specify a startsec as 0? I feel like if it exits quickly, that's a sign there's a problem.

FuhuXia: When I worked on https://github.com/GSA/datagov-deploy/commit/b8103136100eb687b2cb52e32431ce0fe6ecc3e4, I remembered supervisor generated an error complaining it exited too quickly even everything else was fine. We can double check to see if it is still the case for inventory and if `startsecs=0` is still needed.

CodeShtuff: Thank you both. I like the idea of utilizing the default parameter until/if confirmed that a change from the default is even necessary for catalog. Just pushed that change and will confirm that the checks come back clean.

woodt (approved):
Now using default startsecs.  Looks good to me.

PR #2277, opened 2020-10-13
Activate /data.json URL for inventory

avdata99:
Related to [Multi#481](https://github.com/GSA/datagov-ckan-multi/issues/481#issuecomment-707759757)

This PR activate /data.json URL for inventory


avdata99:
Closing since we should create data.json as the user does. We don't want to open this URL

PR #2308, opened 2020-10-19
Remove user/group from supervisord config for catalog

woodt:
Otherwise, we get permission denied:

```
  File "/usr/lib/ckan-new/src/ckan/ckan/lib/i18n.py", line 348, in _build_js_translation
    with open(dest_filename, u'w') as f:
IOError: [Errno 13] Permission denied: u'/usr/lib/ckan-new/src/ckan/ckan/public/base/i18n/am.js'
```


CodeShtuff (approved):
Looks good, removal of these added values should correct perm issue we are seeing.

Issue #2248, opened 2020-10-01
Test and validate the inventory-next switchover plan


FuhuXia:
## Acceptance criteria
- [x] Given I follow the steps described in the plan
  WHEN I visit inventory.data.gov/api/action/status_show
  THEN I see CKAN 2.8 in the response.
- [x] Given I follow the steps described in the plan
  WHEN I visit inventory.data.gov/api/action/datastore_search?resource_id= ([or datastore tests are complete](https://docs.ckan.org/en/2.8/maintaining/datastore.html#test-the-set-up))
  THEN I see the Per Diem API data.

- [x] GIVEN I run the rollback steps described in the plan
    WHEN I visit inventory.data.gov/api/action/status_show
  THEN I see CKAN 2.5 in the response.
- [x] GIVEN I run the rollback steps described in the plan
    WHEN I visit inventory.data.gov/api/action/datastore_search?resource_id=
  THEN I see the Per Diem API data.

## Dry Run
Before the switchover, we have these conditions met:

1. **New INVENTORY** DB RDS is ready, recovered from a recent snapshot of existing Production **INVENTORY** DB.
   **New DATASTORE** DB RDS is ready, recovered from a recent snapshot of existing Production **DATASTORE** DB.
2. New SOLR is ready.
3. New **inventory-next instances** are ready.
4. [House clean work](https://github.com/GSA/datagov-deploy/issues/2758) has been done on Inventory user table.   
5. Dry run has been done and verified on the inventory-next instances, and results met expectations. 

## Switchover plan
1. Take down inventory.data.gov (with maintenance message).
2. Work with FCS to take final **INVENTORY** DB snapshot, and restore it to the **new INVENTORY** Multi-AZ DB. 
    Work with FCS to take final **DATASTORE** DB snapshot, and restore it to the **new DATASTORE** Multi-AZ DB.
    Clean all S3 files under path `/datagov/inventory-next/`:
`aws s3 rm s3://[bucket-name]/datagov/inventory-next/ --recursive`
 Then copy over all content from `/datagov/inventory/`:
`aws s3 cp s3://[bucket-name]/datagov/inventory/ s3://[bucket-name]/datagov/inventory-next/ --recursive`
It will take a few minutes.
3. Run `ckan db upgrade` on **inventory-next instances**.
4. Drop table `related_dataset` first, then drop table `related`. (#2247)
5. Run datastore set-permissions script.
```
[...]

PR #2315, opened 2020-10-21
Inventory-next uses existing RDS snapshot

FuhuXia:
changed DB credentials in Ansible variables to use existing RDS snapshot rather than a new RDS. This will facilitate production switchover.


adborden:
Would you please double check that:
a) we're using a random password
b) we're using a separate db application user from the db admin user

FuhuXia:
> Would you please double check that:
> a) we're using a random password
> b) we're using a separate db application user from the db admin user

@adborden We can make a separate GH issue to address these two points, and have the issue resolved before or after CKAN 2.5 to 2.8 switchover. The purpose of this PR is to have a smooth switchover. The RDS is a snapshot of current production, I think we should to keep user/pass same to minimize any manual DB/Ansible playbook change.

adborden:
I see. In that case I've added these as [post-launch tasks](https://github.com/GSA/datagov-deploy/issues/2248).

jbrown-xentity (approved):

PR #2323, opened 2020-10-22
[hotfix][catalog-next] configuration for production

adborden:
https://github.com/GSA/datagov-ckan-multi/issues/298

catalog-next configuration and secrets. I'm doing this as a hotfix because it
contains vault changes. Since this only includes changes to
inventories/production and only effects catalog_next variables, there's
essentially no effect to hosts. This can be seen with the
ansible-inventory-diff tool, the only change is adding of catalog_next
variables which aren't applied to any hosts.


adborden:
Nevermind, I'm just going to keep the changes together on develop.

PR #2327, opened 2020-10-23
Add cron job to clean harvets logs in worker-next

avdata99:
Related to [Multi#453](https://github.com/GSA/datagov-ckan-multi/issues/453)

Activate the `ckan-clean-harvest-logs` task for worker-next

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/tasks/worker.yml
+++ b/ansible/roles/software/ckan/catalog/ckan-app/tasks/worker.yml
@@ -36,7 +36,7 @@
     minute: "30"
     hour: "2"
     user: root
-    disabled: "{{ not crons_enabled or not catalog_enable_clean_harvest_logs }}"
+    disabled: "{{ not catalog_enable_clean_harvest_logs }}"
     state: present
   tags: ['cron']
 
--- a/ansible/roles/software/ckan/catalog/ckan-app/tasks/worker.yml
+++ b/ansible/roles/software/ckan/catalog/ckan-app/tasks/worker.yml
@@ -36,8 +36,8 @@
     minute: "30"
     hour: "2"
     user: root
-    disabled: "{{ not crons_enabled }}"
-    state: {{ catalog_enable_clean_harvest_logs | ternary('present', 'absent') }}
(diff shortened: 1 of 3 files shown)


ansible/roles/software/ckan/catalog/ckan-app/tasks/worker.yml
```diff
@@ -28,6 +28,18 @@
...
+- name: schedule ckan-clean-harvest-logs cron
+  cron:
+    name: "clean harvest logs in db"
+    cron_file: ckan
+    job: supervisorctl start ckan-clean-harvest-logs > /dev/null
+    minute: "30"
+    hour: "2"
+    user: root
+    disabled: "{{ not crons_enabled }}"
+    state: {{ catalog_enable_clean_harvest_logs | ternary('present', 'absent') }}
```
FuhuXia: Can we do:
```
disabled: {{ not crons_enabled and not catalog_enable_clean_harvest_logs }}
state: present
```

I feel this way logic is easier.

avdata99: I [just changed](https://github.com/GSA/datagov-deploy/pull/2327/commits/7faf22e9b3e4f04d1fce6fe939625bd17776a5e3) @FuhuXia. I use `or` instead `and` because I think that's is the right way here.

FuhuXia: 👍  
Sorry for the and/or mistake.

FuhuXia (approved):

ansible/roles/software/ckan/catalog/ckan-app/tasks/worker.yml
```diff
@@ -28,6 +28,18 @@
     - usr/bin/celery-mem-check.sh
   when: catalog_ckan_worker_type == 'qa_worker'
 
+- name: schedule ckan-clean-harvest-logs cron
+  cron:
+    name: "clean harvest logs in db"
+    cron_file: ckan
```
adborden, line 34: You'll have to give this file a different name. The next task will overwrite /etc/cron.d/ckan and this change will be lost.

FYI, if we had idempotency checks, this would've been caught by molecule.

PR #2330, opened 2020-10-23
use next s3 storage path

FuhuXia:
Use different inventory-next S3 storage path from inventory v1.


adborden (approved):
Inspected vault locally.

jbrown-xentity (approved):

PR #2335, opened 2020-10-23
[redis] fix redis-next

adborden:
catalog-next and inventory-next are using the same redis instance in staging and
production but with different passwords/settings, which doesn't make sense.


jbrown-xentity (approved):
Looks good

PR #2336, opened 2020-10-23
Include PyPI in onboarding/offboarding

adborden:


jbrown-xentity (approved):
Do we want to specify a project here, or just leave blank to do a general review?

adborden:
@jbrown-xentity not sure what you mean? https://pypi.org/project/ckanext-dcat-usmetadata/ is specified and it's currently the only project. Unfortunately pypi doesn't support teams so there's not an easy way to see all data.gov projects. I'm looking at creating a datagov user, but I feel like that might not be the right solution.

jbrown-xentity:
> 
> 
> @jbrown-xentity not sure what you mean? https://pypi.org/project/ckanext-dcat-usmetadata/ is specified and it's currently the only project. Unfortunately pypi doesn't support teams so there's not an easy way to see all data.gov projects. I'm looking at creating a datagov user, but I feel like that might not be the right solution.

I assumed there were other projects in pypi. For this change it's fine that dcat-usmetadata is the only one, however that makes me sad that we don't have any others registered...

adborden:
Planned as part of the forks work: https://github.com/gsa/datagov-ckan-multi/issues/224

PR #2358, opened 2020-10-29
[hotfix][postfix] update sending address (develop)

adborden:
Backmerge https://github.com/GSA/datagov-deploy/pull/2356


jbrown-xentity (approved):
Examined vault locally

Issue #2364, opened 2020-10-29
Inventory Ansible role does not restart supervisor web service after a deploy


adborden:
## How to reproduce

1. Apply the inventory playbook `ansible-playbook inventory.yml`
1. From an inventory machine, check supervisor `ssh inventory1d.dev-ocsit.bsp.gsa.gov sudo supervisorctl restart gunicorn-web-app`

## Expected behavior

Supervisor web service is running for less than a minute

## Actual behavior

```
gunicorn-web-app                 RUNNING    pid 19511, uptime 7 days, 8:09:48
```

jbrown-xentity:
Verified on inventory-next in staging. After running the site playbook, we got the following output (hours after site had been run):

```
$ sudo supervisorctl status
beaker-cache-cleanup        STOPPED     Not started
ckan                                    RUNNING    pid 1502, uptime 2:23:29
```

PR #2366, opened 2020-10-29
[inventory] update supervisor handler

adborden:
https://github.com/GSA/datagov-deploy/issues/2364

Now that we're using supervisor to run ckan (gunicorn), anything that effects
the ckan process should trigger a reload of the ckan process (supervisor).

Apache no longer cares about anything ckan, so it's pretty much a one-to-one
swap of the handlers.

For example, any change to production.ini, to the virtual environment, the
.env, supervisor configs, the ckan start script... apache doesn't care about
any of these, but supervisor does.

I also moved the supervisor service start to the end... no point instarting
supervisor until after the application code is installed (thinking about
first-time install).

### DIFF
--- a/ansible/roles/software/ckan/inventory/handlers/main.yml
+++ b/ansible/roles/software/ckan/inventory/handlers/main.yml
@@ -3,8 +3,11 @@
   service: name=apache2 state=restarted
   when: datagov_in_service
 
-- name: reload supervisor
-  command: supervisorctl reload
+- name: restart supervisor
+  service: name=supervisor state=restarted
+
+- name: restart ckan
+  command: supervisorctl restart ckan
   when: datagov_in_service
   tags:
     - skip_ansible_lint
--- a/ansible/roles/software/ckan/inventory/handlers/main.yml
+++ b/ansible/roles/software/ckan/inventory/handlers/main.yml
@@ -12,6 +12,6 @@
   command: supervisorctl restart ckan
   when:
     - datagov_in_service
-    - inventory-next
+    - inventory_next
   tags:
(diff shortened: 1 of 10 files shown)


adborden:
I'd like to do a little testing in sandbox and make sure this is the right way to manage supervisor config changes. If we're happy with it, let's update all our supervisor roles.

jbrown-xentity (approved):

ansible/roles/software/ckan/inventory/handlers/main.yml
```diff
@@ -3,8 +3,13 @@
   service: name=apache2 state=restarted
   when: datagov_in_service
 
-- name: reload supervisor
-  command: supervisorctl reload
+- name: reread supervisor
```
adborden: Minor, but I would leave this as `reload` for consistency with other services. The fact that we're doing reread instead of reload is just a supervisor implementation detail detail. When tasks are notifying supervisor, all they're saying is "supervisor, you need to reload your configuration files", so "reload supervisor" is a better name IMHO.

ansible/roles/software/ckan/inventory/handlers/main.yml
```diff
@@ -3,8 +3,13 @@
...
+  command: supervisorctl reread
```
adborden: In your catalog-app PR you're calling reread && update. Don't you also want to do that here? For example, since we're renaming the supervisor process gunicorn-web-app to ckan, we want supervisor to apply those changes (removing gunicorn-web-app and adding ckan). I though the update command was necessary to apply it.

Issue #2272, opened 2020-10-09
Inventory-next production environment ready


FuhuXia:
## User Story

Now that inventory-next has been tested in the staging environment, complete the following set up steps to get inventory-next running on production for dry-runs and final switchover.

## Details/tasks
- [x] [Request DNS](https://github.com/GSA/datagov-deploy/wiki/DNS) for inventory-next.data.gov (ticket to BSP)
- [x] [Request SSL certificate](https://github.com/GSA/datagov-deploy/wiki/TLS-SSL-Certificates) for inventory-next.data.gov (ticket to BSP)
- [x] New Load Balancer and/or Netscaler configuration in BSP (ticket to BSP)
- [x] New stand-alone redis instances ~~provisioned with TLS encryption~~( ticket to BSP)
- [x] New Solr instances #2284
- [x] New CKAN DB and Datastore DB loaded with recent production snapshots
- [x] Deploy inventory-next in production #2285

## Acceptance Criteria

- [x] GIVEN inventory-next has been deployed to production \
  AND I spoofed set up has been competed \
  AND set up has been competed \
  WHEN I browse https://inventory-next.data.gov \
  THEN I see the inventory home page \
  OR browse request causes CKAN error and new ticket is created to resolve.

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
[comment]: # "Our SSP says 'The Data.gov team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.' so please don't remove this section without care."
The deployment will use the usual processes for staging/production rollouts, and not introduce any new process or components where security problems are likely to appear.

FuhuXia:
RITM0778598 was created for DNS
RITM0778768 was created for SSL certificate
RITM0778842 was created for NetScaler configuration

FuhuXia:
RITM0778598 DNS ticket was rejected due to improper request. DNS wiki updated.
RITM0778768 SSL certificate ticket: We were given an unchanged cert file. They are working in it to correct it.
RITM0778842 NetScaler ticket was completed.

FuhuXia:
DNS: new ticket RITM0782441 was created.

SSL certificate: the new cert is generated and verified. RITM0782462 is created to request install the the new cert on NetScaler.

FuhuXia:
✅  DNS: RITM0782441 is complete. inventory-next.data.gov is ready.

[...]

PR #2377, opened 2020-11-02
add production inventory-next RDS vars

FuhuXia:
part of https://github.com/GSA/datagov-deploy/issues/2272


jbrown-xentity:
Validated this runs to completion on production (with `datagov_in_service` set to `false`). Still need to finalize a few RDS variables once BSP is done with the provisioning...

FuhuXia:
use `ckan_try_01`, `datastore_try_01`, `datapusher_try_01` as ckan database names, `datastore_ro_try_01` as datastore readonly user, so we can verify the story on a fresh database without any migration work. Will manually run `CREATE database` on the RDS snapshots.

pjsharpe07 (approved):

PR #2412, opened 2020-11-10
Add credentials to next environments

avdata99:
Related to [Multi#480](https://github.com/GSA/datagov-ckan-multi/issues/480)

This PR adds AWS S3 credentials to catalog-next

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
@@ -240,6 +240,11 @@ ckanext.geodatagov.metrics_csv.aws_storage_path = gsa/catalog-next/metrics/
 ckanext.geodatagov.s3sitemap.aws_storage_path = gsa/catalog-next/sitemap/
 ckanext.geodatagov.jsonlexport.aws_storage_path = gsa/catalog-next/jsonl/
 
+ckanext.geodatagov.aws_use_ami_role = true
+ckanext.geodatagov.aws_bucket_name = {{ catalog_bucket_name }}
+ckanext.geodatagov.aws_access_key_id = _placeholder
+ckanext.geodatagov.aws_secret_access_key = _placeholder
+
 # DataGovTheme settings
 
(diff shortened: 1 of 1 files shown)


jbrown-xentity:
What are these credentials for? I looked in geodatagov, but the only instance I can see is [here](https://github.com/GSA/ckanext-geodatagov/blob/master/ckanext/geodatagov/commands.py#L1348), which would not work with this PR.

avdata99:
@jbrown-xentity to [upload the sitemap to S3](https://github.com/GSA/ckanext-geodatagov/blob/73fd9299092d7b33342bfc746c1def38389753f7/ckanext/geodatagov/commands.py#L1128) this is required.

We need to upload the sitemap that we already expose in the [robots file](https://catalog-next.data.gov/robots.txt)
Maybe we need to expose this only for production to avoid to override the production sitemap with staging and sandbox environment

jbrown-xentity:
I believe what was confusing to me was that the only thing that's necessary is the `ckanext.geodatagov.s3sitemap.aws_storage_path` value, the others don't even exist in the code base...

avdata99:
I see we [use the credential in the `get_s3_bucket` function](https://github.com/GSA/ckanext-geodatagov/blob/master/ckanext/geodatagov/commands.py#L1348:L1355).
This function [is required in `sitemap-to-s3` command](https://github.com/GSA/ckanext-geodatagov/blob/master/ckanext/geodatagov/commands.py#L1128:L1128)

Maybe I'm missing something @jbrown-xentity. I think we need these credentials to upload to S3

FuhuXia:
yes, these are need to upload to s3. we also need another one `ckanext.s3sitemap.aws_use_ami_role`.

avdata99:
@FuhuXia this config is already part of the PR

thejuliekramer:
@FuhuXia @jbrown-xentity  we are not sure if we need to add the folder for next manually to the bucket to S3? If so can you give me access the buckets in the Amazon console?

FuhuXia:
No need to add the folder/path manually. S3 "creates" it if it does not exist.

avdata99:
So @FuhuXia, do you think this is ready to merge?

FuhuXia (approved):

PR #2414, opened 2020-11-10
Feature/s3 inventory next

jbrown-xentity:
Utilizing a different CKAN extension for inventory2.8 in conjunction with https://github.com/GSA/datagov-ckan-multi/issues/488.

To be tested and iterated with on sandbox before being ready for review.

### DIFF
--- a/ansible/inventories/sandbox/group_vars/inventory-next/vars.yml
+++ b/ansible/inventories/sandbox/group_vars/inventory-next/vars.yml
@@ -35,7 +35,7 @@ newrelic_enabled: false
 
 ckanext_s3_name: ckanext-cloudstorage
 ckanext_s3_repo: https://github.com/TkTech/ckanext-cloudstorage
-ckanext_s3_req: None
+ckanext_s3_req: false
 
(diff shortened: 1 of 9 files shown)


adborden:
If we're replacing the extension, can we do it in the application via requirements.txt instead of the ansible level? We can't write an integration test unless the extension is included in inventory-app.

jbrown-xentity:
After testing integration, we are no closer to knowing if this extension will work for us than when we started. The documentation is poor, and we are unsure what the various objects are supposed to contain. I am going to hold off removing this draft PR, but this may need to wait for further discussion and/or support.

FuhuXia:
Tested ckanext-cloudstorage on local. Things are working when provided with aws `key` and `secret`. I assume it will work with `IAM` too. Changes need:

1. Freeze urllib3==1.24.3
2. Code modified to support custom path. 
3. Manually make public of the uploaded s3 object after each resource.

jbrown-xentity:
Related to https://github.com/GSA/inventory-app/pull/155, one or the other will be utilized moving forward.

jbrown-xentity:
We are moving forward with local implementation and testing with inventory-app. Will be modifying this PR to remove all s3 modifications for inventory-app...

Issue #2410, opened 2020-11-09
Solr Not Working in Production Catalog-Next


thejuliekramer:
The /harvest /organization and /dataset endpoints using solr are returning 404 messages

## How to reproduce

1. Go to https://catalog-next.data.gov/api/3/action/package_search?q=a
2. Go to https://catalog-next.data.gov/harvest
3. Go to https://catalog-next.data.gov/organization

## Expected behavior

Pages should load without error

## Actual behavior

<img width="1282" alt="Screen Shot 2020-11-09 at 10 48 01 AM" src="https://user-images.githubusercontent.com/7378770/98570709-578a4a00-2279-11eb-9511-40019c95bbfd.png">

The solr error 

<img width="1071" alt="Screen Shot 2020-11-09 at 10 45 15 AM" src="https://user-images.githubusercontent.com/7378770/98571164-db443680-2279-11eb-8ff1-e7a333a0b009.png">

Production vs sandbox `package_search` call: 

![shot-2020-11-09 13 48-1900x860](https://user-images.githubusercontent.com/3237309/98571517-5bc36100-2293-11eb-95c3-ad7a7cd65147.png)

thejuliekramer:
@adborden we updated the solr url in the environment to be pointing to the master solr url and not the replica - the solr replica can't connect to the master solr - we aren't sure what the long term solution is here. Should we update datagov deploy or are we pointing to the replica intentionally and missing some other configuration?

adborden:
It looks like we're accessing the wrong solr instance here. It should be datagov-solr2p<strong>-v2</strong> but in the logs I'm seeing the catalog-classic solr host datagov-solr2p.

avdata99:
@adborden @thejuliekramer I think this PR will do the job: https://github.com/GSA/datagov-deploy/pull/2419/files

PR #2419, opened 2020-11-11
Fix from catalog-next to solr-v2

avdata99:
Related to #2410 

This PR fix connection from catalog-next on production to solr-next


jbrown-xentity (approved):

adborden:
@avdata99 I think the right fix is to just delete these files so that the group_vars kick in. The group vars look correct and are more maintainable since you don't need to add additional host_vars each time we add a new web host.

adborden:
I created a new PR https://github.com/GSA/datagov-deploy/pull/2428

Issue #2125, opened 2020-09-09
HSTS system-wide


jbrown-xentity:
_Please keep any sensitive details in [Google Drive](https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit#heading=h.tka5idg3rc5)._

**Date of report: 9/7/2020** 
**Severity: Medium**
**Due date: 12/6/2020** 

- [ ] Analysis has been performed and an issue has been linked to address other occurrences for this class of vulnerability\* (_link_)

_\* When a finding is identified, we create two issues. One to address the specific instance identified in the report. The other is to identify and address all other occurrences of this vulnerability within the application._

FuhuXia:
should be done after https://github.com/GSA/datagov-deploy/issues/570 is complete.

jbrown-xentity:
Latest scan confirms complete

FuhuXia:
The original Nessus was complaining missing HSTS on the individual host level. I think we still have to address it, otherwise it will come back in another scan.

Issue #570, opened 2019-02-06
Configure web instances to listen on HTTPS


adborden:
data.gov sites redirect HTTP -> HTTPS* but the load balancer talks to the hosts on HTTP. HTTPS would provide an extra layer of security within AWS.

This would involve a few steps because of BSP.
1. Configure apache/nginx to serve the site on both HTTP and HTTPS.
1. Open a ticket with BSP to update all the load balancers to talk to instances over HTTPS.
1. Configure HTTP to 301 redirect to HTTPS so its not possible to connect over HTTP.

This could be done for staging, then production.

_* I'm not exactly sure where this is happening, at CloudFront, load balancer or at NetScaler._

adborden:
CRM and Dashboard were completed as part of [RITM0598393](https://gsa.service-now.com/sp/?id=form&table=sc_req_item&filter=&sys_id=de1c654adb64f7440e1d368d7c961919&v=).

adborden:
No longer applies to component/static since we moved to Federalist.

FuhuXia:
Will do a survey on following Netscaler/Load Balancer entries and request BSP to change protocol from all HTTP to HTTPS. This change also means currently used ip address in the URL need to be changed to hostnames.

Prod:
```
wp-bsp.data.gov 		(www.data.gov)
catalog-bsp.data.gov 		(catalog.data.gov)
admin-catalog-bsp.data.gov	(admin-catalog.data.gov)
catalog-next.data.gov
admin-catalog-next.data.gov
inventory-bsp.data.gov		(inventory.data.gov)
dashboard-bsp.data.gov		(labs.data.gov/dashboard/)
```
Staging:
```
wordpress-datagov.dev-ocsit.bsp.gsa.gov
catalog-datagov.dev-ocsit.bsp.gsa.gov
admin-catalog-datagov.dev-ocsit.bsp.gsa.gov
catalog-next-datagov.dev-ocsit.bsp.gsa.gov
admin-catalog-next-datagov.dev-ocsit.bsp.gsa.gov
inventory-datagov.dev-ocsit.bsp.gsa.gov
inventory-next-datagov.dev-ocsit.bsp.gsa.gov
dashboard-datagov.dev-ocsit.bsp.gsa.gov
```

PR #2426, opened 2020-11-13
add HSTS header to catalog and inventory

FuhuXia:
For https://github.com/GSA/datagov-deploy/issues/2125


adborden:
@FuhuXia should we take on https://github.com/GSA/datagov-deploy/issues/570 first?

Since the instances themselves listen on 80, having the HSTS header set might cause some problems behind netscaler. It's probably fine, I think only web browsers care about HSTS and they are going through https but we should just tackle the https issue to get it over with and avoid headaches. If something does go wrong, it's not easy to fix because of the long lifespan of HSTS.

FuhuXia:
#570 is handled. We are ready to push the HSTS change. 
@adborden On Netscaler level we have HSTS in place already. This change is just silence NESSUS scan. It won't affect client browser or risk wrong HSTS configuration.

pjsharpe07 (approved):

PR #2429, opened 2020-11-13
Apache rewrite rollback

jbrown-xentity:
Rolling back changes that broke redirects for admin side of the application (changes tested and validated on production server).

Issue #2083, opened 2020-08-26
catalog / redirects to HTTP


adborden:
We should be redirecting to HTTPS to avoid the extra redirect request.

## How to reproduce

1. `curl -v https://catalog.data.gov/ > /dev/null`

## Expected behavior

302 https://catalog.data.gov/dataset

## Actual behavior

302 http://catalog.data.gov/dataset ->
301 https://catalog.data.gov/dataset

adborden:
I think this also affects inventory, but haven't confirmed.

mogul:
This may be resolved with CKAN-next. Will check on it when we are upgraded!

avdata99:
The catalog-next in the sandbox already include the expected redirection

<pre>
$ curl -v catalog-next.sandbox.datagov.us > /dev/null

> GET / HTTP/1.1
> Host: catalog-next.sandbox.datagov.us
> User-Agent: curl/7.68.0
> Accept: */*
<b>< HTTP/1.1 302 Found</b>

< Date: Mon, 09 Nov 2020 10:51:48 GMT
< Content-Type: text/plain; charset=utf8
< Transfer-Encoding: chunked
< Connection: keep-alive
< Server: gunicorn/19.10.0
<b>< Location: https://catalog-next.sandbox.datagov.us/dataset</b>
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: POST, PUT, GET, DELETE, OPTIONS
< Access-Control-Allow-Headers: X-CKAN-API-KEY, Authorization, Content-Type
< Referrer-Policy: origin
</pre>

I'm not able to found where this behaviour is defined but it seems to work in the sandbox

New production is not ready

<pre>
curl -v catalog-next.data.gov > /dev/null
> GET / HTTP/1.1
> Host: catalog-next.data.gov
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
<b>< HTTP/1.1 302 Found : Moved Temporarily</b>
<b>< Location: https://catalog-next.data.gov/</b>
< Connection: close
< Cache-Control: no-cache
< Pragma: no-cache
</pre>

adborden:
~I think this needs to be fixed within the CloudFront distribution or at the Netscaler level.~

adborden:
[...]

PR #2433, opened 2020-11-16
Add redirect from / to /dataset

avdata99:
Related to #2083 

After the [previous PR failed](https://github.com/GSA/datagov-deploy/pull/2429/files) this is a new try

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
@@ -72,11 +72,16 @@ SSLEngine on
     RewriteCond %{HTTP_REFERER}    !{{saml2_idp_entry}} [NC]
     RewriteCond %{HTTP_COOKIE}     !auth_tkt
     RewriteRule ^(.*)$ {{ url_readonly }}$1 [R,L]
+
 {% elif catalog_ckan_readwrite_configuration == 'readonly' %}
     RewriteRule   "^(/en)?/user/login"  "{{ url_writable }}/user/login" [R,L]
     RewriteRule   "^(/en)?/user/_logout"  "{{ url_writable }}/user/_logout"  [R,L]
     RewriteRule   "^(/en)?/user/logged_in"  "{{ url_writable }}/user/logged_in"  [R,L]
     RewriteRule   "^(/en)?/login_generic"  "{{ url_writable }}/login_generic"  [R,L]
+
+    RewriteCond %{HTTP_REFERER} !{{saml2_idp_entry}} [NC]	
+    RewriteRule ^(\/?)$ /dataset [R]
+
 {% endif %}
 
     RewriteRule ^/api/rest/dataset$ api/action/package_search [R]
@@ -91,4 +96,5 @@ SSLEngine on
 
     RewriteRule ^/api/3/action/resource_search$ api/action/package_search [R]
     RewriteRule ^/api/action/resource_search$ api/action/package_search [R]
+    
 </VirtualHost>
--- a/ansible/smoke.yml
+++ b/ansible/smoke.yml
@@ -158,7 +158,7 @@
     - name: assert only one redirection to https and /dataset
       block:
         - name: look for expected redirection
-          shell: "curl -L --max-redirs 1 -v http://catalog-next.data.gov > /dev/null"
+          shell: "curl -L --max-redirs 1 -v https://catalog-next.data.gov > /dev/null"
           register: response
           failed_when: response.rc != 0
             
--- a/ansible/smoke.yml
+++ b/ansible/smoke.yml
@@ -151,26 +151,17 @@
       run_once: true
       when: smoke_enable_cdn_checks | default(True)
 
-- name: catalog-next.data.gov /dataset redirection
+- name: catalog-next.data.gov tests
   hosts: all
   gather_facts: false
   tasks:
     - name: assert only one redirection to https and /dataset
-      vars:
-        url: http://catalog-next.data.gov
       block:
         - name: look for expected redirection
-          uri:
-            url: "{{ url }}"
-            status_code: 302
-            follow_redirects: all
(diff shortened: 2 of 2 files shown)


jbrown-xentity:
@avdata99 can we add a smoke test to cover the issue that arose the last time? Try to hit the admin site, and see if it redirects appropriately?

avdata99:
I added [a smoke test](https://github.com/GSA/datagov-deploy/pull/2433/commits/abd9309d085810cf2273a06fa7c4076a6e767166) @jbrown-xentity 
Do you think is enough?

jbrown-xentity (approved):
This is what I was thinking, but using production in every smoke test could cause [issues if catalog is down/out on production](https://github.com/GSA/datagov-deploy/issues/2350#issuecomment-718091862) when rolling to sandbox/staging. Not a huge deal, and should not occur often, but something to note. This also doesn't give us validation at the staging level, since it's testing production...

avdata99:
I see @jbrown-xentity. 
I also noted that the `maximum_redirection` I'm using is for [`win_uri` module](https://docs.ansible.com/ansible/2.8/modules/win_uri_module.html) and not for [`uri` module](https://docs.ansible.com/ansible/2.8/modules/uri_module.html).
Did you know why?
And I expect a fail because of this but the test is green. Why is not failing?

jbrown-xentity:
This test isn't even run as part of circleci, but will cause the playbook to fail (we'll see when we run manually or on jenkins). It's a helpful catch for something that is difficult to validate in molecule...
As for the `maximum redirection` flag, I'm not sure...

ansible/smoke.yml
```diff
@@ -150,3 +150,28 @@
...
+  tasks:
+    - name: assert only one redirection to https and /dataset
+      vars:
+        url: http://catalog-next.data.gov
+      block:
+        - name: look for expected redirection
+          uri:
+            url: "{{ url }}"
+            status_code: 302
+            follow_redirects: all
```
adborden: This won't work. If you follow redirects, then you'll never see a status code of 302. You want `follow_redirect: none`.

ansible/smoke.yml
```diff
@@ -150,3 +150,28 @@
...
       run_once: true
       when: smoke_enable_cdn_checks | default(True)
+
+- name: catalog-next.data.gov /dataset redirection
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: assert only one redirection to https and /dataset
+      vars:
+        url: http://catalog-next.data.gov
```
adborden: According to the [bug](https://github.com/GSA/datagov-deploy/issues/2083), we should be testing **https** not http.

We don't really care if http://catalog.data.gov is two hops, as along as it's secure.

Why are are we testing catalog-next.data.gov and not catalog.data.gov?

ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
```diff
@@ -72,11 +72,16 @@ SSLEngine on
...
     RewriteRule ^(.*)$ {{ url_readonly }}$1 [R,L]
+
 {% elif catalog_ckan_readwrite_configuration == 'readonly' %}
     RewriteRule   "^(/en)?/user/login"  "{{ url_writable }}/user/login" [R,L]
     RewriteRule   "^(/en)?/user/_logout"  "{{ url_writable }}/user/_logout"  [R,L]
     RewriteRule   "^(/en)?/user/logged_in"  "{{ url_writable }}/user/logged_in"  [R,L]
     RewriteRule   "^(/en)?/login_generic"  "{{ url_writable }}/login_generic"  [R,L]
+
+    RewriteCond %{HTTP_REFERER} !{{saml2_idp_entry}} [NC]	
+    RewriteRule ^(\/?)$ /dataset [R]
```
adborden, line 83: I'm not really sure how this solves the issue. The problem is that `https://catalog.data.gov/` is redirecting to `http://catalog.data.gov/dataset`, so this new rule doesn't change that.

ansible/smoke.yml
```diff
@@ -150,3 +150,28 @@
...
+            register: response
+        - name: assert redirection
+          assert:
+            that: response.url == 'https://catalog-next.data.gov/dataset'
+            fail_msg: |
+              We expect a redirection to HTTPS and to /dataset
+      delegate_to: localhost
+      become: false
+      run_once: true
+      when: ckan_uses_gunicorn | default(True)
```
adborden: Why ckan_uses_gunicorn? ckan_uses_gunicorn variable is not defined for all hosts, so it will likely always end up with `default(True)`.

ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
avdata99: @adborden, we are now redirecting to `/dataset` at CKAN level [inside DataGovTheme](https://github.com/GSA/ckanext-datagovtheme/blob/master/ckanext/datagovtheme/plugin.py#L81:L81).
Issue #2083 mention we have two redirections and we expect just one. 
I assume we have the _https_ redirection in Apache and the second one, inside CKAN.
I think we need to add this redirection inside apache like we already [are doing at the inventory](https://github.com/GSA/datagov-deploy/blob/2083_redirect2/ansible/roles/software/ckan/inventory/templates/etc/apache2/sites-enabled/ckan.conf#L45:L45)

ansible/smoke.yml
avdata99: Since this task is delegated to `localhost` and `run_once` is present, the `when` line is not needed, I just delete it.

ansible/smoke.yml
avdata99: I fixed the URL, I used _http_ by mistake.
I also changed to test just the production URL.

adborden:
@avdata99 is this still needed?

avdata99:
No @adborden, this PR is not useful. Closing

PR #2447, opened 2020-11-18
Redirect tracking POST to writable URL

avdata99:
Related to [datagovcatalog#20](https://github.com/GSA/ckanext-datagovcatalog/issues/20)

This PR redirects `/_tracking` POST to writable URL

Tested manually in production for catalog-next and is not working (not sure why)
We are [changing JS code](https://github.com/GSA/datagov-deploy/blob/20_fix_tracking_POST/ansible/roles/software/ckan/fix-ckan-tracking/tasks/main.yml#L5) from a task in ansible to point to the right URL.

Maybe a redirection is better than these code-changes
Note: [status 307](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) should preserve POST and avoid changing it to GET. We are redirecting HTTP to HTTPS with a 302 redirection which transforms the POST requests into a GET one. This GET request gives us 404 response. Maybe we also need to fix the HTTP to HTTPS redirect

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
@@ -72,6 +72,7 @@
     RewriteRule   "^(/en)?/user/_logout"  "{{ url_writable }}/user/_logout"  [R,L]
     RewriteRule   "^(/en)?/user/logged_in"  "{{ url_writable }}/user/logged_in"  [R,L]
     RewriteRule   "^(/en)?/login_generic"  "{{ url_writable }}/login_generic"  [R,L]
+    RewriteRule   "^/_tracking$" "{{ url_writable }}/_tracking" [R=307,L]
 {% endif %}
 
     RewriteRule ^/api/rest/dataset$ api/action/package_search [R]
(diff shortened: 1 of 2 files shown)


thejuliekramer (approved):

avdata99:
307 preserve method but 302 redirect to GET which give us a 404 error

<pre>
curl -L -v https://catalog-next.data.gov/_tracking -d "url/dataset/department-for-the-aging-dfta-geriatric-mental-health-contracted-providers&type=page" > /dev/null

> <b>POST</b> /_tracking HTTP/1.1
< HTTP/1.1 307 Temporary Redirect
< Server: Apache
< Location: http://admin-catalog-next.data.gov/_tracking

> <b>POST</b> /_tracking HTTP/1.1
</pre>

FuhuXia (approved):

adborden:
@avdata99 does this mean we can remove the [fix-ckan-tracking role](https://github.com/GSA/datagov-deploy/tree/develop/ansible/roles/software/ckan/fix-ckan-tracking)?

avdata99:
> @avdata99 does this mean we can remove the [fix-ckan-tracking role](https://github.com/GSA/datagov-deploy/tree/develop/ansible/roles/software/ckan/fix-ckan-tracking)?

I think this is a better and simpler way to handle these redirections. If this works we can remove this role.
I thought to do that but not sure about this redirection.

avdata99:
I added a new PR [here](https://github.com/GSA/datagov-deploy/pull/2456) @adborden

PR #2451, opened 2020-11-19
Optimize Gunicorn

avdata99:
Related to [Multi#513](https://github.com/GSA/datagov-ckan-multi/issues/513)

This PR:
 - Define properly the number of workers for Gunicorn (after a load testing failed)
 - Allows set the size of sqlalchemy pools. Not in use but ready to use in the future if needed.

### DIFF
--- a/ansible/inventories/production/group_vars/catalog-next/vars.yml
+++ b/ansible/inventories/production/group_vars/catalog-next/vars.yml
@@ -1,7 +1,7 @@
 ---
 app_repo: https://github.com/GSA/catalog.data.gov
 catalog_ckan_app_version: master
-
+catalog_ckan_wsgi_processes: 8
 catalog_ckan_apache_server_alias:
   - catalog.data.gov
 catalog_ckan_apache_server_name: "{{ catalog_host_public_next }}"
@@ -25,10 +25,9 @@ catalog_db_name: "{{ catalog_ckan_db_name }}"
 catalog_db_pass: "{{ catalog_ckan_db_pass }}"
 catalog_db_user: "{{ catalog_ckan_db_user }}"
 
-sqlalchemy_pool_size: 15
-sqlalchemy_max_overflow: 25
-datastore_sqlalchemy_pool_size: 15
-datastore_sqlalchemy_max_overflow: 25
+# Before increasing pool_size check gunicorn workers
+# sqlalchemy_pool_size: 15
+# sqlalchemy_max_overflow: 25
 
 ckan_catalog_next: true
 ckan_instance_secret: "{{ catalog_next_ckan_instance_secret }}"
--- a/ansible/inventories/production/group_vars/catalog-next/vars.yml
+++ b/ansible/inventories/production/group_vars/catalog-next/vars.yml
@@ -25,6 +25,11 @@ catalog_db_name: "{{ catalog_ckan_db_name }}"
 catalog_db_pass: "{{ catalog_ckan_db_pass }}"
 catalog_db_user: "{{ catalog_ckan_db_user }}"
 
+sqlalchemy_pool_size: 15
+sqlalchemy_max_overflow: 25
+datastore_sqlalchemy_pool_size: 15
+datastore_sqlalchemy_max_overflow: 25
+
 ckan_catalog_next: true
 ckan_instance_secret: "{{ catalog_next_ckan_instance_secret }}"
 ckan_instance_uuid: "{{ catalog_next_ckan_instance_uuid }}"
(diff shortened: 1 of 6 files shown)


ansible/inventories/production/group_vars/catalog-next/vars.yml
```diff
@@ -25,6 +25,12 @@ catalog_db_name: "{{ catalog_ckan_db_name }}"
 catalog_db_pass: "{{ catalog_ckan_db_pass }}"
 catalog_db_user: "{{ catalog_ckan_db_user }}"
 
+override_sqlalchemy_pools: true
+sqlalchemy_pool_size: 15
+sqlalchemy_max_overflow: 25
+datastore_sqlalchemy_pool_size: 15
+datastore_sqlalchemy_max_overflow: 25
```
jbrown-xentity: datastore isn't enabled on catalog (or shouldn't be), so those 2 lines shouldn't be necessary, here or in the production.ini file...

jbrown-xentity (commented):
Might want to remove the datastore components...

ansible/inventories/production/group_vars/catalog-next/vars.yml
avdata99: Just fixed by removing this settings @jbrown-xentity. Thanks

jbrown-xentity (approved):

thejuliekramer (approved):

ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
```diff
@@ -1,5 +1,5 @@
 [program:catalog-web]
-command=/etc/ckan/server_start.sh -b localhost:5000 --proxy-allow-from="127.0.0.1"
+command=/etc/ckan/server_start.sh -b localhost:5000 --workers=$(( 2 * `cat /proc/cpuinfo | grep 'core id' | wc -l` + 1 )) --proxy-allow-from="127.0.0.1"
```
adborden: Let's use an Ansible variable here so that we can tune it. 2 x CPU cores is only the recommended number to start at.

avdata99: Fixed [here](https://github.com/GSA/datagov-deploy/pull/2451/commits/13988b9753b4cec3084a965ae558f37abd9a40e8) @adborden

ansible/roles/software/ckan/catalog/ckan-app/templates/supervisord_gunicorn.conf.j2
```diff
@@ -1,5 +1,5 @@
 [program:catalog-web]
-command=/etc/ckan/server_start.sh -b localhost:5000 --workers=$(( 2 * `cat /proc/cpuinfo | grep 'core id' | wc -l` + 1 )) --proxy-allow-from="127.0.0.1"
+command=/etc/ckan/server_start.sh -b localhost:5000 --workers={{ gunicorn_workers|default(1) }} --proxy-allow-from="127.0.0.1"
```
adborden: This should go into the role defaults. Since this variable is very similar to [catalog_ckan_wsgi_processes](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml#L17), let's rename that variable and use it for both apache-wsgi and for gunicorn.

avdata99: @adborden I remove `gunicorn_workers` and use `catalog_ckan_wsgi_processes` for all. Default value is 2 * CPUs + 1. Removed the production value. All environments will use this default:
https://github.com/GSA/datagov-deploy/pull/2451/commits/c997c236313871b5f0e6bd365e1e6cc419cd241a

ansible/inventories/production/group_vars/all/vars.yml
```diff
@@ -20,7 +20,6 @@ catalog_ckan_service_url: https://{{ catalog_host_public }}
 catalog_ckan_admin_service_url: https://{{ catalog_host_admin }}
 catalog_ckan_solr_port: "8983"
 catalog_ckan_solr_host: datagov-solr1p.prod-ocsit.bsp.gsa.gov
-catalog_ckan_wsgi_processes: 6
```
adborden: Are you sure you want to change this? On catalog-classic, I think we're running m4.2xlarge with 8 vcpus and 32G of ram. That would default to 17 ckan processes which tend to be memory hungry. Are you sure that catalog-classic can handle that?

avdata99: I'm not sure which is the right number of workers for each context. Maybe a load test could help to define it.
For production, I leave `6` as it was before.
Considering this, I added a `4` for stagging and `2` for sandbox.

adborden (approved):

PR #2456, opened 2020-11-19
Remove unused role and test 307 /_tracking redirection

avdata99:
Related to [datagovcatalog#20](https://github.com/GSA/ckanext-datagovcatalog/issues/20)
Continue fro [previous PR](https://github.com/GSA/datagov-deploy/pull/2447)

This PR:
 - Remove the `fix-ckan-tracking` role
 - Add smoke test for `/_tracking` redirection


ansible/smoke.yml
```diff
@@ -150,3 +150,28 @@
...
+      block:
+        - name: send POST
+          uri:
+            url: https://catalog.data.gov/_tracking
+            method: POST
+            status_code: 307
+            follow_redirects: none
+      delegate_to: localhost
+      become: false
+      run_once: true
```
adborden, line 171: Please assert the Location header is correct. You can add a `register` and then `assert` that the location is correct (pointing to admin-catalog.data.gov).
There's an example in [datagov-web](https://github.com/GSA/datagov-deploy/blob/develop/ansible/datagov-web.yml#L50-L71) (ignore the loop).

adborden (approved):
👍  this works for catalog-classic?

ansible/smoke.yml
avdata99: I added the assertions: https://github.com/GSA/datagov-deploy/pull/2456/commits/c142c03cc23e671c2cd86863c45a563f83092609

avdata99:
>  this works for catalog-classic?

@adborden this will work for catalog-classic and also for catalog-next after the migration

PR #2497, opened 2020-11-30
Disable Dan Mayol's access

mogul:
Offboarding per https://github.com/GSA/datagov-deploy/issues/2374


jbrown-xentity:
I believe we should be removing the users now @mogul .
See [this code block](https://github.com/GSA/datagov-deploy-jumpbox/blob/master/tasks/main.yml#L31-L69) and [these tests](https://github.com/GSA/datagov-deploy-jumpbox/blob/master/molecule/users/tests/test_default.py). Maybe you could update the comments accordingly so others don't get confused?

mogul:
@jbrown-xentity Oh I missed that change! Yes, I've made a new commit on the branch to remove Dan's entry and the comments.

jbrown-xentity (approved):

Issue #900, opened 2019-08-20
Ansible smoke tests with uri module should use validate_certs=yes


adborden:
We use smoke tests in our playbooks to assert that web services are up and running using the [Ansible uri module](https://docs.ansible.com/ansible/latest/modules/uri_module.html). The module fails with:
```
Request failed: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:727)>
```

This is unexpected because the certificate is a GSA signed certificate installed to the well-known `/etc/ssl/certs/ca-certificates.crt` location.

## How to reproduce

1. `ansible -m uri -a 'url=https://crmweb1d.dev-ocsit.bsp.gsa.gov status_code=200' crmweb1d.dev-ocsit.bsp.gsa.gov`

## Expected behavior

Success, 200 status code.

## Actual behavior

SSL certificate verify failure.

nickumia-reisys:
No more Ansible,
- https://github.com/GSA/data.gov/issues/3710

Our current smoke tests use `curl` which automatically fails if [certificates are expires (or otherwise bad)](https://stackoverflow.com/a/62115626).

PR #2502, opened 2020-12-02
Bug/fix smoke tracking

jbrown-xentity:
To get around #2485. Smoke test fails for tracking url on production, this changes to only test on production (and clean up a few variables).


ansible/inventories/production/group_vars/all/vars.yml
```diff
@@ -25,6 +25,7 @@ catalog_pycsw_db_host: "{{ vault_catalog_pycsw_db_host }}"
 catalog_pycsw_db_name: "{{ vault_catalog_pycsw_db_name }}"
 catalog_pycsw_db_pass: "{{ vault_catalog_pycsw_db_pass }}"
 catalog_pycsw_db_user: "{{ vault_catalog_pycsw_db_user }}"
+catalog_validate_redirect: true
```
adborden: Since this is specific to the smoke playbook, I would prefix it with that. Also, follow the existing convention from the `smoke_enable_cdn_checks` variable.
```suggestion
smoke_catalog_tracking_check: true
```

adborden (approved):
As I'm looking at this, I'm realizing we want this enabled on staging too. But since we're testing the apache config, these smoke tests should be part of the catalog playbook, not smoke. That way these could be testing each catalog-web host in staging and production. The way this is written now, it's not easy to do that.

ansible/inventories/production/group_vars/all/vars.yml
jbrown-xentity: I think we want to pivot then to something more like this newest commit, using different variables to hit the tracking URL and then when the instance is readonly then validate that the redirect is working...

jbrown-xentity:
I think the latest commit should be the more correct fix, however I get an error when I test on staging.
Not sure why there would be an SSL error on the public route, everything on staging looks good to me.
@adborden any ideas?
```
TASK [send POST] ****************************************************************************************************************************************************************************
fatal: [catalog-web1d.dev-ocsit.bsp.gsa.gov -> localhost]: FAILED! => {"changed": false, "content": "", "elapsed": 0, "msg": "Status code was -1 and not [307]: Request failed: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852)>", "redirected": false, "status": -1, "url": "https://catalog-datagov.dev-ocsit.bsp.gsa.gov/_tracking"}
```

adborden:
Unfortunately you have to skip ssl validation on staging https://github.com/GSA/datagov-deploy/issues/900

ansible/catalog-web.yml
```diff
@@ -77,6 +77,57 @@
       when: datagov_in_service | default(true)
       tags:
         - smoke
+      
+    - name: check /_tracking first redirection
+      block:
+        - name: send POST
+          uri:
+            url: "{{ url_readonly }}/_tracking"
```
adborden: I think this makes more sense at the host level, not the service level.

```suggestion
            url: "{{ ansible_fqdn }}/_tracking"
```

jbrown-xentity: Yes, I just thought this was getting tripped up on the cert and tried the more public URL. Now that I know about #900, I will roll back to this and utilize the `validate_certs` field in the uri module...

jbrown-xentity:
Latest tests are passing, and changes made show working in staging. Plan to squash and then merge once final tests pass...

PR #2506, opened 2020-12-03
[hotfix][catalog-next] POSTLAUNCH redirect traffic

adborden:
To be merged after launch https://github.com/GSA/datagov-ckan-multi/issues/400

- After launch, any requests to catalog-next.data.gov should be redirected to
  catalog.data.gov.


adborden:
Not necessary, we already removed catalog-next.data.gov from allowed hosts, so we are just returning 404 for the domain.

PR #2507, opened 2020-12-03
[hotfix][catalog-next] PRELAUNCH

adborden:
To be merged just before launch https://github.com/GSA/datagov-ckan-multi/issues/400

- Serve requests for catalog.data.gov


jbrown-xentity (approved):
Approving ahead of launch...

Issue #2220, opened 2020-09-29
inventory.data.gov uses login.gov for authn


mogul:
## User Story

As an organization data admin, I want to be able to access the latest version of the inventory app in a secure way so that I know that my data is secure.

## Acceptance Criteria

- [x] GIVEN login.gov is setup on inventory.data.gov \
  WHEN a user goes to https://inventory-next.data.gov/user/login [or staging equivalent] \
  THEN the user can login to login.gov \
  AND the user is redirected back to inventory-next.data.gov with the correct permissions

## Background
Should utilize [ckanext-saml2auth ckan2.8 extension](https://github.com/keitaroinc/ckanext-saml2auth/tree/ckan-2.8).
Catalog implemented this already, see [PR](https://github.com/GSA/catalog.data.gov/pull/198/files) and [configuration](https://github.com/GSA/catalog.data.gov/pull/199/files). Tests are in work, and should also be implemented; this should work locally if configured properly.
May also need to implement [change in login route similar to catalog](https://github.com/GSA/ckanext-datagovtheme/pull/80/files).
Will require change/addition in login.gov sandbox configuration for inventory, check catalog configuration.

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
[comment]: # "Our SSP says 'The Data.gov team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.' so please don't remove this section without care."
See SSP for user management details

jbrown-xentity:
Paired with @woodt today. Went over the current status of https://github.com/GSA/datagov-ckan-multi/issues/348. To catch up to where catalog work currently is, we should:

- Setup an application on login.gov sandbox similar to [catalog-next](https://dashboard.int.identitysandbox.gov/service_providers/1129)
- Create draft PR's similar to the work for [catalog-next](https://github.com/GSA/datagov-ckan-multi/issues/348#issuecomment-706270960) (can use the current ckanext-saml2 branch, no need to create a new one).

Note that this will still be in a failing state, but we did work through the [encryption issue](https://github.com/GSA/datagov-ckan-multi/issues/348#issuecomment-715453866) (that setting is done in login.gov, not on a per-request basis as is normally done for saml2).

FuhuXia:
App `inventory-next-sandbox` has been created on login.gov sandbox. 
[...]

PR #2519, opened 2020-12-07
Inventory next saml login.gov

jbrown-xentity:
Related to https://github.com/GSA/inventory-app/pull/178 and https://github.com/GSA/datagov-deploy/issues/2220. Being tested on sandbox for final components, last pending issue is redirect after login


jbrown-xentity:
To manually setup the DB, we needed to run the DB setup command for ckanext-saml:
`paster --plugin=ckanext-saml2 saml2 create -c /etc/ckan/production.ini`
This is now working, with a small issue of the redirect to `/dataset` after logging in not working (if you go there manually, you are logged in). Will continue looking at the fix, while marking as ready for review...

Issue #2553, opened 2020-12-10
Monit service for supervisor/gunicorn ckan processes


adborden:
## User Story

In order to keep ckan running without alerts requiring human intervention, data.gov operator wants a monit service to monitor supervisor/gunicorn ckan processes.

## Acceptance Criteria
[ACs should be clearly demoable/verifiable whenever possible. Try specifying them using [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).]
- [ ] GIVEN I run the catalog-web playbook
  WHEN a ckan process uses more than `monit_ckan_mem_threshold`% of system memory
  THEN monit restarts the ckan processes.

## Background
[Bulk of the monit work](https://github.com/GSA/datagov-deploy/pull/1947). We switched from apache wsgi to supervisor/gunicorn, so any memory should be monitored at the supervisor job process level.

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
None

## Sketch
The ansible role we're using has a [process_by_name](https://github.com/pgolm/ansible-role-monit/blob/7207b640cbb363c9c141d2eacb9938998416e6c2/templates/monitor.j2#L6) type to key off of a unique process name instead of a pid file. You might need to tweak the supervisor job's [process_name](http://supervisord.org/configuration.html#program-x-section-values) so you have something unique to match against. gunicorn also has a [proc_name](https://docs.gunicorn.org/en/latest/settings.html#proc-name) option. The start/stop commands should use `supervisorctl` start/stop.

gunicorn runs multiple child worker processes, so be careful about which process you're monitoring, or if monit can monitor the process group instead. @FuhuXia did the original monit work and might have ideas on how to make this work.

Also consider setting gunicorn's [max-requests](https://docs.gunicorn.org/en/latest/settings.html#max-requests)/[max-requests-jitter](https://docs.gunicorn.org/en/latest/settings.html#max-requests-jitter) options. Monit should kick in if memory get's out of hand, but if gunicorn is recycling processes regularly, then I would expect memory to be reasonable without monit stopping all the workers.

nickumia-reisys:
No more Ansible.  No more supervisor.
- https://github.com/GSA/data.gov/issues/3710

There is no need for monitoring ckan memory on cloud.gov.  If the ckan app becomes unhealthy (crashes), cloud.gov restarts it.  We also have restarts every 15 mins (although not for this reason).  Either way, I don't believe the motivation for this feature still exists.

PR #2554, opened 2020-12-10
[catalog-next] add max-requests

adborden:
https://github.com/GSA/datagov-deploy/issues/2553

Recycle gunicorn workers every 5000-6000 requests. These numbers are a rough
guess to start. Previously this is unset, so workers are never recycled.

Assuming we want to recycle processes every ~60 minutes, with a total,
consistent site load of 2000 req/min and 4 web instances each with 6 processes
(on production):

    60m * 2000r/m / (4*6) = 5000r

And then we use a 20% jitter to stagger restarts, effectively restarting every
60-72 minutes assuming the load never changes.


jbrown-xentity (approved):
Do we need to do this for inventory too @adborden?

adborden:
@jbrown-xentity couldn't hurt for inventory.

Issue #2312, opened 2020-10-20
Nessus Scan 10/20/2020


pjsharpe07:
_Please keep any sensitive details in [Google Drive](https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit#heading=h.it2rh2xc99f1)._

**Date of report:** 10/20/2020
**Severity:** Medium
**Due date:** 01/20/2021

_Due date is based on severity and described in RA-5. 15-days for Critical, 30-days for High, and 90-days for Moderate and lower._

- [x] Analysis has been performed and an issue has been linked to address other occurrences for this class of vulnerability\* (_link_)

_\* When a finding is identified, we create two issues. One to address the specific instance identified in the report. The other is to identify and address all other occurrences of this vulnerability within the application._

### Brief description
[Link to Google Drive](https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit#heading=h.it2rh2xc99f1)

mogul:
@pjsharpe07 can you link here to the upstream issue about making it possible to upgrade without this workaround?

pjsharpe07:
[link to current upstream issue](https://github.com/robertdebock/ansible-role-tomcat/issues/22)

PR #2559, opened 2020-12-14
Tomcat version upgrade

pjsharpe07:
Manual work still needed to be done on servers. Will check staging first


jbrown-xentity (approved):
Don't forget, will need to remove the folder and re-install (bumping version variable does not force a re-install)...

pjsharpe07:
Part of [this work](https://github.com/GSA/datagov-deploy/issues/2312)

PR #2600, opened 2021-01-04
[common] custom production shell prompt

adborden:
Remind operators that they are on a production system.


jbrown-xentity (approved):
Will be a great indicator, thanks!

adborden:
FYI, this only applies to the ubuntu user. I think I started a change for jumpbox...

Issue #2597, opened 2021-01-04
wp-bsp.data.gov is open to public


FuhuXia:
wp-bsp.data.gov is supposed to be used internally as Origin server by CloudFront, not accessible by general public. But instead, it receives too much traffic.  88% of traffic in /var/log/nginx/error.log is targeted to wp-bsp.data.gov.
```
root@wordpressweb1p:/var/log/nginx# grep wp-bsp error.log|wc -l
4136
root@wordpressweb1p:/var/log/nginx# grep -v wp-bsp error.log|wc -l
564
```

## How to reproduce

open wp-bsp.data.gov in browser.

## Expected behavior

see 404 page.

## Actual behavior

see same content as www.data.gov.

FuhuXia:
We can follow [the fix](https://github.com/GSA/datagov-deploy/issues/1491) we did for catalog-bsp.data.gov

PR #2636, opened 2021-01-11
[wordpress] remove wp-bsp from allowed hostnames

adborden:
https://github.com/GSA/datagov-deploy/issues/2597

WP should return a 404 for wp-bsp domain instead of serving traffic. This helps
reduce load on WP and improves SEO.


adborden:
🤔  okay I will fix that. I think 404 is more correct than a 403 in this case.

FuhuXia:
404 changes merged on wordpress and dashboard. Please create release tags and update ansible/requirements.yml.

adborden:
@FuhuXia PR is updated, ready for another look.

FuhuXia (approved):

Issue #174, opened 2017-05-26
 Add New Relic application agents


kvuppala:
When this story is up for the next sprint: We think it should be broken up into one issue per app.

---
As an application developer, I need my application configured with New Relic so that I can can monitor and alert across all hosts.

## Acceptance criteria
- [ ] New Relic is added as an application dependency to app repo
- [ ] Application hosts report as a single application
- [ ] Only production and staging are enabled with staging labeled separately
- [ ] Configuration lives in application
- [ ] Apdex alerts are configured

## Details

We currently have language-specific playbooks to deploy and configure New Relic for applications. However, this has some deficiencies. The platform has to know details about the application, environments, app names, configuration settings. The platform has to know where to install hooks within the application (wsgi, worker scripts, etc). The platform might install conflicting versions of dependencies, outside of application pinned dependencies. The platform installing New Relic might cause an extra restart, prolonging any service disruption during deploy.

kvuppala:
All front applications now have New Relic applications installed, Solr and Harvester agents are pending.

kvuppala:
@dano-reisys We also need to group instances so each application metrics are consolidated.

adborden:
Updated description.

mogul:
We think this one probably should be broken up into one issue per app.

nickumia-reisys:
This is super old and written for a different infrastructure.  For cloud.gov specifically, NR is installed on catalog, inventory and dashboard.  The only place we haven't implemented this is our Solr deployment, 
- https://github.com/GSA/data.gov/issues/3956

PR #2651, opened 2021-01-15
[catalog][inventory] fix NR reporting

adborden:
This fixes a bug where staging and sandbox instances are reporting to the
production APM app.

This updates the New Relic configration for APM in a consistent way so that
environment is a separate app (or falls into the "unknown" bucket).

The New Relic variables are removed from the catalog and inventory roles and
specified through the dot env file. This way the roles don't need to know about
the details of New Relic, there's a standard interface (environment variables)
to configure NR.


pjsharpe07:
Mentioning [the original ticket](https://github.com/GSA/datagov-deploy/issues/174). Leaving this so we'll update the original ticket status to reflect that we don't want to report on staging statistics.

jbrown-xentity (approved):
Cleaned up a bunch, thanks!

PR #2662, opened 2021-01-21
Fix resource_search endpoint

avdata99:
The `resource_search` API endpoint is been redirected to `package_search`
It seems this is an error and we need this endpoint to finish our[ load testing tool](https://github.com/GSA/datagov-load-testing/blob/advanced_test/locust/advanced.py#L114-L132)

The [commit](https://github.com/GSA/datagov-deploy/commit/b32de59301d5198e44e669b4eb03cb4fd8730be2) that applies this change is not clear about why. Probably a mistake.

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
@@ -99,7 +99,4 @@
 
     RewriteRule ^/api/action/current_package_list_with_resources$ api/action/package_search [R]
     RewriteRule ^/api/3/action/current_package_list_with_resources$ api/3/action/package_search [R]
-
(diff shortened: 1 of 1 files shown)


FuhuXia:
We have a lot more resources than datasets. resource_search api requests tends to consume too much system resource. If this change go to production, we will see more incidents of web instance OOM.

avdata99:
> We have a lot more resources than datasets. resource_search api requests tends to consume too much system resource. If this change go to production, we will see more incidents of web instance OOM.

@FuhuXia Ok. Closing this PR and avoid resource load testing

adborden:
@avdata99 can you open a bug for this? We should not be disabling resource_search.

adborden:
BTW, this would be good to put in the commit message, so we're not guessing as to why it was removed.

> We have a lot more resources than datasets. resource_search api requests tends to consume too much system resource. If this change go to production, we will see more incidents of web instance OOM.

avdata99:
New issue (#2664) and PR (#2665) created to keep track of this @adborden @FuhuXia

Issue #2575, opened 2020-12-21
Inventory authentication extension


jbrown-xentity:
## User Story

As an inventory developer, I want a well defined authentication extension so that authentication is well documented, can be changed, and the logic is isolated for inventory.

## Acceptance Criteria
- [x] GIVEN inventory web is running \
  and the ckanext-inventory_auth extension is enabled \
  WHEN an anonymous user requests any page other than a resource \
  THEN the user is redirected to the login page (or equivalent)

## Background
From research on [inventory anonymous user seeing dataset pages in inventory-2.8](https://github.com/GSA/datagov-deploy/issues/2314)

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
[comment]: # "Our SSP says 'The Data.gov team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.' so please don't remove this section without care."
[Any security concerns that might be implicated in the change. "None" is OK, just be explicit here!]
Should contain tests to verify that pages and API is inaccessible

## Sketch
We will want to create and store this extension in the [inventory-app](https://github.com/gsa/inventory-app).
Utilize the [CKAN documentation](https://docs.ckan.org/en/2.9/extensions/) and [previous extension work](https://github.com/GSA/ckanext-datagovcatalog).
The original implementation of this is [here](https://github.com/GSA/USMetadata/blob/master/ckanext/usmetadata/plugin.py#L581-L593), but behavior was inconsistent and some metadata was discoverable that was not intended. Use this as a starting point, but may need to implement access limitations for all CKAN actions except for resource viewing/download access.

chris-macdermaid:
Ran into this open CKAN issue: 
Chained auth functions lose anonymous access
https://github.com/ckan/ckan/issues/5751

We can create a auth chain in the new extension for logged in users, but because of the open issue this won't work to allow anonymous users to see resources. There is an active open [PR](https://github.com/ckan/ckan/pull/5756) to fix this.

pjsharpe07:
Remaining steps for this ticket:

- [x] Rename [current extension](https://github.com/GSA/inventory-app/tree/feature/ckanext-iauth/ckanext-datagov_iauthfunctions) to be `datagov_inventory` or something similar
- [x] Confirm all changes in [ckan's default auth](https://github.com/ckan/ckan/tree/2.8/ckan/logic/auth) behavior is what we want and make any remaining changes/edits
[...]

PR #2681, opened 2021-01-25
Update add datagov_inventory extension

chris-macdermaid:
This is related to https://github.com/GSA/datagov-deploy/issues/2575


ansible/roles/software/ckan/inventory/templates/etc/ckan/production.ini
```diff
@@ -109,8 +110,12 @@ ckan.redis.url = {{ inventory_ckan_redis_url}}
 #		Add ``resource_proxy`` to enable resorce proxying and get around the
 #		same origin policy
 # ckan.plugins = usmetadata datajson datastore datapusher stats text_view recline_view googleanalyticsbasic saml2 dcat_usmetadata
+{% if inventory_next %}
+ckan.plugins = {{ (inventory_next_ckan_plugins_default + inventory_ckan_plugins_default + inventory_ckan_plugins_additional) | join(' ') }}
+{% else %}
```
pjsharpe07: I think we'll want to close this if statement. Similar to [this](https://github.com/GSA/datagov-deploy/blob/develop/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2#L32-L47)

chris-macdermaid: I was able to get rid of the if/else.

ansible/roles/software/ckan/inventory/templates/etc/ckan/production.ini
```diff
@@ -85,8 +85,9 @@ ckan.auth.user_create_organizations = false
 ckan.auth.user_delete_groups = false
 ckan.auth.user_delete_organizations = false
 ckan.auth.create_user_via_api = false
-ckan.auth.create_user_via_web = true
+ckan.auth.create_user_via_web = false
```
jbrown-xentity, line 88: @FuhuXia can you confirm with the saml extension this is the correct setting? Or do we want something else?

FuhuXia: As of now, saml2 extension is using a separated production.ini template. So any time we make changes to production.ini file, the same changes need to be applied to ansible/roles/software/ckan/inventory/templates/inventory-next/etc_ckan_production.ini.j2

adborden: So, we should update inventory-next/etc_ckan_production.ini.j2 in this PR?

FuhuXia: Yes, although I doubt this will have any effect on saml login behaviour. But for consistency, we need to keep two production.j2 synchronized.

chris-macdermaid: Updated ansible/roles/software/ckan/inventory/templates/inventory-next/etc_ckan_production.ini.j2 with the changes.

adborden (approved):

PR #2683, opened 2021-01-26
Allow to ask PIV card

avdata99:
This PR allows defining the required requested contexts (e.g. PIV card) for saml2 login


ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2
```diff
@@ -237,6 +237,10 @@ ckanext.saml2auth.want_response_signed=false
 ckanext.saml2auth.want_assertions_signed=false
 ckanext.saml2auth.want_assertions_or_response_signed=true
 ckanext.saml2auth.enable_ckan_internal_login=false
+{% if catalog_ckan_saml2_require_piv %}
```
adborden: Instead of creating a boolean value, create a variable for the string value. That keeps it simple while allowing flexibility.

```
# vars.yml
catalog_ckan_saml2_requested_authn_context:
  - http://idmanagement.gov/ns/assurance/aal/3?hspd12=true

# production.ini.j2
{% if catalog_ckan_saml2_requested_authn_context | length > 0 %}
ckanext.saml2auth.requested_authn_context = {{ catalog_ckan_saml2_requested_authn_context | join(' ') }}
{% endif %}

Comment (avdata99) [reply ansible/roles/software/ckan/catalog/ckan-app/templates/catalog-next/etc_ckan_production.ini.j2]: I just updated the PR to add this @adborden

**Review by adborden — APPROVED** [review]

**Comment by adborden** [comment]
Oh, one thing. You should add a default to ckan-app/defaults/main.yml

```
catalog_ckan_saml2_requested_authn_context: []
```

For environments where the variable isn't defined, we want to default to an empty array.

Issue #1311, opened 2020-01-28
CloudFront cache hits are zero


adborden:
In our prod AWS account, looking at the cloudfront statistics for catalog.data.gov, we see zero cache hits.

<img width="1012" alt="Screen Shot 2020-01-27 at 2 45 01 PM" src="https://user-images.githubusercontent.com/509703/73290571-871d0980-41b3-11ea-9e99-3ebc0fdb8424.png">

## How to reproduce

1. Request a dataset

        curl -v 'https://catalog.data.gov/dataset/range-wild-horse-and-burro-territory-feature-layer-0d6a0' 2>&1  > /dev/null | grep -i x-cache
        < x-cache: Miss from cloudfront

1. Request a dataset

        curl -v 'https://catalog.data.gov/dataset/range-wild-horse-and-burro-territory-feature-layer-0d6a0' 2>&1  > /dev/null | grep -i x-cache
        < x-cache: Miss from cloudfront

## Expected behavior

Second request shows a cache hit.

## Actual behavior

Both requests show a cache miss.

adborden:
In 2017 there was an explicit decision to set no-cache headers in Catalog. https://github.com/GSA/ckan/commit/6bba96029181ba112e44fa82b60daf2e20fc0ddc

mogul:
The comment above that was added in this commit, which has a lot more logic for when to cache! 
https://github.com/GSA/ckan/commit/0a0b595a9a099de11a56b2e96604a9856b944e1a

Maybe there's a bug there which means that the "else" clause is getting hit more often than it should?

jbrown-xentity:
catalog-app local development currently broken, breaking [here](https://github.com/GSA/ckanext-datagovtheme/blob/master/ckanext/datagovtheme/helpers.py#L105). Seems to not know what the URL is, but the config option is set... [Current PR](https://github.com/GSA/catalog-app/pull/118) (not ready for review)

jbrown-xentity:
catalog-app is fixed. Tracked error to occurring here: https://github.com/GSA/ckan/blob/datagov/ckan/lib/base.py#L188
There must be a new inclusion (extension?) that causes this if statement to succeed every time and cause the cache to break. Could consider removing this specific if statement to re-enable caching.

jbrown-xentity:
[...]

PR #2707, opened 2021-02-01
Add Cache-Control max-age to catalog-next

avdata99:
Related to [Multi#449](https://github.com/gsa/datagov-ckan-multi/issues/449)

As mentioned in #1311 CKAN fails to handle max-age and the [fix](https://github.com/ckan/ckan/pull/4781/files) is not backported to CKAN 2.8.
Maybe we can force this in apache configuration

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
+++ b/ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
@@ -56,7 +56,7 @@
     CustomLog {{ catalog_ckan_access_log }} combined
 
     # Cache
-    Header add Cache-Control "max-age={{ catalog_ckan_cache_control_max_age }}"
+    Header set Cache-Control "max-age={{ catalog_ckan_cache_control_max_age }}"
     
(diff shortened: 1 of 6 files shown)


ansible/roles/software/ckan/catalog/ckan-app/templates/etc_apache2_sites-enabled_ckan.conf.j2
```diff
@@ -55,6 +55,9 @@
     ErrorLog {{ catalog_ckan_error_log }}
     CustomLog {{ catalog_ckan_access_log }} combined
 
+    # Cache
+    Header set Cache-Control "max-age={{ catalog_ckan_cache_control_max_age }}"
```
adborden, line 59: We should let CKAN specify the cache headers. Let's leave this line out.

adborden (changes requested):
@avdata99 we've been down this road before. CKAN should manage all the cache headers. The apache config will go away with cloud.gov and we want a caching solution that works at the application level.

avdata99:
I agree @adborden. CKAN should handle this. Since CKAN is failing to do it now I think this could be a fix
I will look hot to fix CKAN

PR #2722, opened 2021-02-02
[catalog] enable PIV authentication with SAML2

adborden:
Re-enable SAML2 now that PIV is enabled.


adborden:
Re-doing as a hotfix.

adborden:
https://github.com/GSA/datagov-deploy/pull/2724

PR #2776, opened 2021-02-09
Define analytics settings for production

avdata99:
Related to [Multi#405](https://github.com/GSA/datagov-ckan-multi/issues/405)

Define Google Analytics settings for each catalog-next environment

### To validate
Not sure about `googleanalytics.token.filepath` since is not found in analytics extension
We are using it for catalog-classic but is not clear on how it works.

### DIFF
--- a/ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml
+++ b/ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml
@@ -59,8 +59,7 @@ catalog_enable_clean_harvest_logs: false
 catalog_log_dir: /var/log/ckan
 catalog_gunicorn_error_log: "{{ catalog_log_dir }}/gunicorn.log"
 
-googleanalytics_ids:
-  - UA-XXXXXXX-1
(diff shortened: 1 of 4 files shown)


ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml
```diff
@@ -59,6 +59,9 @@ catalog_enable_clean_harvest_logs: false
 catalog_log_dir: /var/log/ckan
 catalog_gunicorn_error_log: "{{ catalog_log_dir }}/gunicorn.log"
 
+googleanalytics_ids:
+  - UA-XXXXXXX-1
```
adborden: This should default to an empty list `[]`.

adborden (approved):

ansible/roles/software/ckan/catalog/ckan-app/defaults/main.yml
avdata99: Fixed

PR #2930, opened 2021-03-03
use inventory.data.gov for inventory-next

FuhuXia:
Use production url `inventory.data.gov` when inventory-next goes live after switchover.
Config change on login.gov Data.gov Inventory app need to go together with this.


jbrown-xentity (approved):
Will our saml configurations break? Do we need to change those on login.gov?

Issue #2931, opened 2021-03-04
Jenkins polling is not detecting changes


adborden:
Last datagov-deploy build was 2/27 87e3c18776e9b726595060ec4def83eb7dda8bbb  https://ci-datagov.mgmt-ocsit.bsp.gsa.gov/job/deploy-ci-platform/lastBuild/

Last master commit was 3/3 431cd9bf3c54935ce00b02202fc544295e549e6f

Polling log says no changes https://ci-datagov.mgmt-ocsit.bsp.gsa.gov/job/deploy-ci-platform/scmPollLog/ but there definitely have been.
```
Started on Mar 3, 2021 11:59:00 PM
Using strategy: Default
[poll] Last Built Revision: Revision 87e3c18776e9b726595060ec4def83eb7dda8bbb (refs/remotes/origin/master)
The recommended git tool is: NONE
No credentials specified
 > git --version # timeout=10
 > git --version # 'git version 2.17.1'
 > git ls-remote -h -- https://github.com/gsa/datagov-deploy.git # timeout=10
Found 62 remote heads on https://github.com/gsa/datagov-deploy.git
Using strategy: Default
[poll] Last Built Revision: Revision 87e3c18776e9b726595060ec4def83eb7dda8bbb (refs/remotes/origin/master)
The recommended git tool is: NONE
No credentials specified
 > git --version # timeout=10
 > git --version # 'git version 2.17.1'
 > git ls-remote -h -- https://github.com/gsa/datagov-deploy.git # timeout=10
Found 62 remote heads on https://github.com/gsa/datagov-deploy.git
Done. Took 0.28 sec
No changes
```

## How to reproduce

Push up a new commit to `master` and look for an update at the URL above.

## Expected behavior

Changes are detected and rolled out.

## Actual behavior

"No changes"

## Sketch
[Notes or a checklist reflecting our understanding of the selected approach]

jbrown-xentity:
I believe https://github.com/GSA/catalog.data.gov/pull/256 did not deploy either; would be interested in pairing to resolve @adborden ...

PR #2946, opened 2021-03-06
[jenkins] fix polling

adborden:
https://github.com/GSA/datagov-deploy/issues/2931

:shrug:

Scratching my head on this. There are several references to Jenkins polling not
detecting changes or building when no changes are detected. Some of these
issues are recent, some not recent. Some have work-arounds, the work-arounds
didn't seem to work. I'm just kind of poking at it and guessing and I think I
have a working configuration.

This updates sandbox to use polling (so we can debug it) in the same way that
production is configured. The best I can figure, it doesn't like having the
dynamic branch specifier based on the branch_name parameter. So let's remove
it. That was originally the start of being able to do release branch builds,
but now we're not planning on doing that.

_Update: spent some more time with @mogul and after trying a few more things and digging through the Jenkins git plugin code, we decided to switch to the multibranch pipeline which uses different code for polling._

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,19 @@
 pipeline {
   agent any
+  options {
+    disableConcurrentBuilds()
+    copyArtifactPermission('*')
+  }
   stages {
     stage('workflow:sandbox') {
-      when { anyOf { environment name: 'DATAGOV_WORKFLOW', value: 'sandbox' } }
+      when {
+        allOf {
+          environment name: 'DATAGOV_WORKFLOW', value: 'sandbox'
+          anyOf {
+            branch 'develop'
+          }
+        }
+      }
       stages {
         stage('build') {
           steps {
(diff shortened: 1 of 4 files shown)


jbrown-xentity (dismissed):

Jenkinsfile
```diff
@@ -1,21 +1,30 @@
...
     stage('workflow:sandbox') {
-      when { anyOf { environment name: 'DATAGOV_WORKFLOW', value: 'sandbox' } }
+      when {
+        allOf {
+          environment name: 'DATAGOV_WORKFLOW', value: 'sandbox'
+          anyOf {
+            branch 'develop'
+          }
+        }
+      }
```
mogul, line 16: Does this stage not require [an `environment` block like the one below](https://github.com/GSA/datagov-deploy/pull/2946/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R50-R53)?

mogul (approved):

Issue #2941, opened 2021-03-05
Remove Inventory Classic Instances


jbrown-xentity:
## User Story

In order to reduce the number of hosts we have to patch and maintain, data.gov team wants some of the inventory-classic FCS resources decommissioned.

## Acceptance Criteria
[ACs should be clearly demoable/verifiable whenever possible. Try specifying them using [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).]
- [x] WHEN I browse to inventory-datagov.dev-ocsit.bsp.gsa.gov \
  THEN I get an unresolved domain error.

## Background
https://github.com/GSA/datagov-deploy/issues/2858

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
[comment]: # "Our SSP says 'The Data.gov team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.' so please don't remove this section without care."
None

## Sketch
- [x] Open FCS ticket to terminate hosts (RITM0820817)
  - inventory1d
  - inventory2d
  - inventory1p
  - inventory2p
  - datagov-solr1d
  - datagov-solr2d
  - datagov-solr1p
  - datagov-solr2p
- [x] Remove above hosts from staging and production hosts file
- [x] Remove above hosts from [System Inventory](https://docs.google.com/spreadsheets/d/1ZYahWuDasufr8VXCHJSlrXHB6-UYk7g2agU87MMKnCk/edit#gid=1724757408) and notify ISSO
- [x] Remove inventory-datagov.dev-ocsit.bsp.gsa.gov Netscaler and DNS record (RITM0821532)
- [x] Remove RDS (RITM0820882)
  - datagov-inventory-96.cy0fc5s0csfm.us-east-1.rds.amazonaws.com
  - datagov-datastore-96.cy0fc5s0csfm.us-east-1.rds.amazonaws.com
  - datagov-inventory.cwyyxljkfrkk.us-east-1.rds.amazonaws.com
  - datagov-datastore-96.cwyyxljkfrkk.us-east-1.rds.amazonaws.com

adborden:
@FuhuXia for RDS, the only thing to remove here is the staging RDS instance for inventory, right? There is no additional RDS instance in production?

FuhuXia:
@adborden Production 2.8 RDS was built from classic RDS snapshots. We still need to remove classic RDS.

adborden:
[...]

PR #2951, opened 2021-03-08
[solr] remove solr-classic replicas

adborden:
https://github.com/GSA/datagov-deploy/issues/2941

We still have catalog-admin which requires the solr-classic primary. The
replicas can be removed.


jbrown-xentity (approved):
Looks good, ready to go with inventory branch change!

Issue #3178, opened 2021-05-03
Automate Inventory publishers.csv deployment


FuhuXia:
## User Story

In order to reduce the manual effort to update publishers in inventory, data.gov team wants the `publishers.csv` checked into the `datagov-deploy` repository and have it auto deployed upon any update.

## Acceptance Criteria
[ACs should be clearly demoable/verifiable whenever possible. Try specifying them using [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).]
- [x] GIVEN I browse to [datagov-deploy repository](https://github.com/GSA/datagov-deploy)\
  THEN I can see `publishers.csv` \
  AND it contains the latest data from [google sheet ](https://docs.google.com/spreadsheets/d/1BWpXWswnnMRaNazMCYrzl4W5JMWpGuyW0Wo2YaNH9jY/edit#gid=0).
- [x] GIVEN a change made in `publishers.csv`\
  WHEN I [add a dataset](https://inventory.data.gov/dataset/new-metadata) in the Inventory app \
  THEN I see the change reflected in the Organization/Publisher dropdown
## Background
[Any helpful contextual notes or links to artifacts/evidence, if needed]

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))
[comment]: # "Our SSP says 'The Data.gov team ensures security implications are considered as part of the agile requirements refinement process by including a section in the issue template used as a basis for new work.' so please don't remove this section without care."
[Any security concerns that might be implicated in the change. "None" is OK, just be explicit here!]

PR #3200, opened 2021-05-05
add publishers.csv into inventory additional-tasks.yml

FuhuXia:
For https://github.com/GSA/datagov-deploy/issues/3178


jbrown-xentity (approved):
Looks good! The section `publishers_result.rc != 0`, is this what gracefully handles errors in sandbox/staging where there are no organizations?

FuhuXia:
> Looks good! The section `publishers_result.rc != 0`, is this what gracefully handles errors in sandbox/staging where there are no organizations?

Not really. `publishers-import `command will gives a warning and move on if no organization found. But in the next task, empty db without a real organization will fail since I need to see at least one  "Updated publishers" result. So this does require developer to manually create a GSA org on an empty (sandbox/staging) DB, as stated in the fail message.

```
- name: Fail when publishers.csv is not loaded
  fail:
    msg: "At least one real orgnization need to be created, such as General Services Administration"
  when: publishers_result.stdout is defined and "Updated publishers" not in publishers_result.stdout
```

PR #3233, opened 2021-05-17
Update inventory-publishers.csv

hkdctol:


FuhuXia:
@hkdctol The new csv does not look right.
Look at what we have before, and look at the new one you brought it, and see if you can have the csv correctly formatted.
[before](https://github.com/GSA/datagov-deploy/blob/master/ansible/files/inventory-publishers.csv) and [after](https://github.com/GSA/datagov-deploy/blob/hkdctol-patch-1/ansible/files/inventory-publishers.csv)

FuhuXia:
replaced by https://github.com/GSA/datagov-deploy/pull/3234

Issue #2970, opened 2021-03-10
Upgrade FGDC2ISO


jbrown-xentity:
_Please keep any sensitive details in [Google Drive](https://docs.google.com/document/d/1rW0VOzfCrjXOI0O1gTAOm225_fTbu5Cjrg7yJTYYBBg/edit#heading=h.o034x9yaisv4)._

**Date of report: 2021-03-08** 
**Severity: Medium**
**Due date: 2021-06-06** 

_Due date is based on severity and described in RA-5. 15-days for Critical, 30-days for High, and 90-days for Moderate and lower._

- [x] Analysis has been performed and an issue has been linked to address other occurrences for this class of vulnerability\* (N/A)

_\* When a finding is identified, we create two issues. One to address the specific instance identified in the report. The other is to identify and address all other occurrences of this vulnerability within the application._

### Brief description
Ideally this application can be deprecated once https://github.com/GSA/datagov-deploy/issues/2917 is done, and this work does not need to be taken on.

FuhuXia:
The PR :point_up_2: contains good info for future tomcat upgrade task.

adborden:
Bumped the version of Tomcat. We already monitor new versions of components, no vulnerability class issue necessary.

PR #3235, opened 2021-05-17
Update tomcat

adborden:
https://github.com/GSA/datagov-deploy/issues/2970

- Bump tomcat version
- Bump tomcat role

Newer versions of the role exist, but 4.6 introduces a "bug" in that it assumes the existence of the
default app (/opt/tomcat/webapps/ROOT) which we explicitly remove.

The latest verison of the role is 5.1 which requires a newer version of Ansible.

We still need to work around the role not supporting an update-in-place.

- [x] After merging, run:

      $ ansible -m file -a 'state=absent path=/opt/tomcat' catalog-fgdc2iso
      $ ansible-playbook fgdc2iso.yml


FuhuXia (approved):

adborden:
Manually deployed to sandbox, staging, and production.

PR #3313, opened 2021-06-28
extend and add ANSIBLE-1300676

FuhuXia:


nickumia-reisys (commented):
Everything looks okay to me, but I don't feel confident for my approval to be "the" approval.

The only comment is that three days were extended to July and one was extended to August.  I guess I don't know the context for this.

FuhuXia:
@nickumia-reisys The one extended to August is because there are other similar Ansible exceptions that are extended to August. We know we are not likely to do anything on them since we have an Ansible 3.x upgrading in July milestone. For others, if nothing changes,  I usually extend them for another two week so we revisit in the next sprint.

nickumia-reisys (approved):
I understand it better now.  When we do review it further, I'd be interested on the future plan.

PR #3459, opened 2021-09-29
Add nginx default upload max size

jbrown-xentity:
Transition to https://github.com/nginxinc/ansible-role-nginx caused change in default `max_body_size` from 64m to 1m. Getting requests from users to bump this back up.

Removes unused defaults from previous nginx deployment.

Did not see `client_max_body_size` support in `nginx_config_main_template` config, only in `nginx_config_http_template` config. `fgdc2iso` app is configured using `nginx_config_http_template` , therefore the edit is done in ansible file. `dashboard` and `wordpress` are configured using a conf file template, so the edits are done in their repos, we add them in by updating the requirement tag.

pending on new tags after https://github.com/GSA/datagov-deploy-dashboard/pull/32 and https://github.com/GSA/datagov-deploy-wordpress/pull/27

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -7,52 +7,52 @@ ignore:
         reason: >-
           solaris_zone module is not used, fix requires ansible 2.9
           https://github.com/gsa/datagov-deploy/issues/893
-        expires: 2021-09-30T00:00:00.000Z
+        expires: 2021-10-14T00:00:00.000Z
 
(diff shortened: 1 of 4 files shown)


FuhuXia:
deploy to sandbox https://ci.sandbox.datagov.us/job/deploy-ci-platform-mb/job/develop/196/, all looks good.

FuhuXia (approved):

PR #3702, opened 2022-02-28
remove wordpress hosts from Jenkins job

FuhuXia:


ansible/inventories/staging/hosts
```diff
@@ -85,4 +85,3 @@ wordpress-web
 wordpress-web-v2
 
 [wordpress-web-v2]
```
nickumia-reisys, line 87: Do you not have to remove this too?

FuhuXia: It is ok to have this line, making it an empty group without any hosts.

nickumia-reisys (approved):
🚢

PR #3886, opened 2022-07-11
Enable egress github action

robert-bryson:


.github/workflows/enable-egress.yml
```diff
@@ -11,6 +11,10 @@ on:
         description: 'Cloud Foundry Space'
         required: true
 
+env:
+  DATAGOV_DIR: datagov
```
nickumia-reisys, line 15: Where is this directory?  I don't see it in main branch nor being added here 😕 Is this specifying the name of the folder that the repo gets cloned to?

nickumia-reisys (commented):
Would it be easier to specify the environment as a matrix, so that the same code doesn't have to be duplicated three times? 
Reference: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations

.github/workflows/enable-egress.yml
robert-bryson: It is used as the path for the [datagov repo](https://github.com/GSA/data.gov/blob/de441e6b2d486410a5e3ce47686e98fb0e38836b/.github/workflows/enable-egress.yml#L30), same as [CG_DIR](https://github.com/GSA/data.gov/blob/de441e6b2d486410a5e3ce47686e98fb0e38836b/.github/workflows/enable-egress.yml#L35), during the execution of the action. They are needed to provide a path so the script is able to reach into the other dir/repo.

robert-bryson:
> Would it be easier to specify the environment as a matrix, so that the same code doesn't have to be duplicated three times? Reference: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations

I believe a matrix in github actions works in the inverse way: provide a matrix of options and run the action with all permutations. We are trying to input a specific option and run an action with just those parameters.

nickumia-reisys (approved):
⛰️

nickumia-reisys:
Linking to 
- https://github.com/GSA/data.gov/issues/3873

PR #3921, opened 2022-08-09
neat, you can do dropdowns now!

robert-bryson:


robert-bryson:
With a good [disable](https://github.com/GSA/data.gov/actions/runs/2828769426) and [enable](https://github.com/GSA/data.gov/runs/7756567596) run, it looks like these dropdowns just work Pretty neat!

jbrown-xentity (approved):

PR #3930, opened 2022-08-18
egress proxy for inventory

robert-bryson:
with https://gsa-tts.slack.com/archives/C2N85536E/p1660773794628549 this might start working.


nickumia-reisys:
This is already part of main, couldn't track down the exact PR that did it, but we have enabled egress on inventory already.

PR #3977, opened 2022-09-28
Duplicate identifier checker

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/3968

Add new helper script that finds datasets with duplicate identifiers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,5 @@ gen-csr-out
 # ansible-galaxy vendored roles
 ansible/roles/vendor
 .vscode/settings.json
+
+output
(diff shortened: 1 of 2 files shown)


Jin-Sun-tts (approved):

jbrown-xentity:
I think this is great, and could be improved to evaluate the type of duplicate that is occurring. For example:
1. Do the dataset `harvest_object_id`'s match?
2. How far apart were they harvested (`metdata_modified` field)?
3. Are the duplicates identical in every way, or are there differences? (this may be difficult with jq to check every field)
4. Does the dataset only exist in solr (use `package_show` to check if it exists outside of SOLR)

But it's definitely a nice, clean, simple script!

nickumia-reisys:
If using VIM, the following command cleans up the output..
```vim
g/^"\x\{8}-\x\{4}-\x\{4}-\x\{4}-\x\{12}"$/normal kJ
```

References:
- https://superuser.com/a/168967
- https://stackoverflow.com/a/20487511

PR #4075, opened 2022-11-21
Test older version of egress-proxy

nickumia-reisys:
Try pinning egress-proxy to old code to see if it works..

References:
- https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit


nickumia-reisys:
From investigation, we found a bug that we (@btylerburton @FuhuXia @nickumia-reisys) created.  Sorry @btylerburton @FuhuXia @nickumia-reisys 
- #4076

PR #4124, opened 2022-12-14
bump action versions

robert-bryson:
Related to
- https://github.com/GSA/data.gov/issues/4005

After thinking about it for a sec, [this](https://github.com/GSA/us-data-federation/pull/68) should probably go here.


nickumia-reisys (approved):
🚯

PR #4157, opened 2023-01-18
Restarts reformat

robert-bryson:
Related to https://github.com/GSA/data.gov/issues/3326

### DIFF
--- a/bin/check-running-job
+++ b/bin/check-running-job
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+function check_running_job {
+    # Input any app name
+    app_to_restart=$1
+
+    # Install jq
+    apt install -y jq
+
+    # Check if deployment is happening, if so ignore
+    guid=$(cf app "$1" --guid)
+    running=$(cf curl "/v3/deployments?status_values=ACTIVE" | jq --arg guid "$guid" ".resources[] | select(.relationships.app.data.guid == \$guid ) | .relationships.app.data.guid")
+    if [ "$running" != "" ]; then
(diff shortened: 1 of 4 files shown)


robert-bryson:
This [should be working](https://github.com/GSA/inventory-app/actions/runs/3989134618) for restarts. I would like to merge this template and then make the related PRs for [inventory](https://github.com/GSA/inventory-app/pull/536) and catalog. Once those are in and working, I'll make the PRs that would effect deploys (the other half of this story).

bin/check-running-job
```diff
@@ -0,0 +1,33 @@
...
+
+    # If there is more than one instance of 'instances', the app is already restarting
+    if [[ $(cf app "$app_to_check" | grep -c '^instances:') -gt 1 ]]; then
+      echo "Deployment or Restart in progress... not doing anything"
+      exit 0
+    fi
+}
+
+# attempting https://stackoverflow.com/a/29967433/20735210
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
```
nickumia-reisys, line 31: Is this only because you are `source`ing this file?

robert-bryson: It's to mimic a python-like `if __name__ is __main__` type thing so I can use it directly as well as in the `restart-app` script.

nickumia-reisys: Fair enough, so only for local testing?

nickumia-reisys (approved):
🤖

jbrown-xentity (approved):

PR #4198, opened 2023-02-07
Fix restart/deploy issue template locations

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/3326

I've been getting a bunch of restart failures, but the issues aren't being created.

image

### DIFF
--- a/.github/deploy_failure.md
+++ b/.github/deploy_failure.md
@@ -1,0 +1,11 @@
+---
+title: Deployment Failure
+labels: bug
+---
+
+Workflow with Issue: {{ workflow }}
+Job Failed: {{ env.GITHUB_JOB }}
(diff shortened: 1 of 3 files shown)


robert-bryson:
Oh duh. Thanks, @nickumia-reisys!

robert-bryson (approved):

robert-bryson:
As a follow up cleanup PR for [catalog](https://github.com/GSA/catalog.data.gov/pull/768) and [inventory](https://github.com/GSA/inventory-app/pull/542). 

I imagine that we would want to point to these for every repo/action we run, right?

jbrown-xentity:
That makes sense. And I might be misunderstanding, but it looks like this PR references the deploy failure template for a restart failure...

PR #4301, opened 2023-05-05
no vendoring for python apps

FuhuXia:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4259

## About

Newer version of python buildpack (1.8.5 and above) is more strict on faulty vendored installs. Instead of fixing what is wrong with our CKAN app vendoring process, it is easier not to use vendored installs. 

## PR TASKS

- [x] The actual code changes.
- [ ] ~Tests written and passed.~
- [ ] ~Any changes to docs?~
- [ ] ~Any dependent PRs needed?~

### DIFF
--- a/.github/workflows/deploy-template.yml
+++ b/.github/workflows/deploy-template.yml
@@ -33,10 +33,6 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


nickumia-reisys (approved):
🥪

PR #4374, opened 2023-06-30
allow staging to use its own allow.acl

FuhuXia:
## About

Sometimes `develop` space is used for different version testing and we could not test on the egress. Why not let staging be able to have its own egress rule.

Copied prod allow.acl to staging, added *.tigbox.com for testing.


nickumia-reisys (approved):
🍅

PR #4394, opened 2023-07-18
remove ckan 2.9.7 from list of ckan matrix versions to test

btylerburton:
Just a white noise PR to remove CKAN version that will always fail from our testing matrix


robert-bryson (approved):
What a great pleasure to approve!

PR #4453, opened 2023-09-06
Retry deployments until something gives

nickumia-reisys:
Related to:
- https://github.com/GSA/data.gov/issues/4261

Changes:
- If a deployment is attempted and fails, retry again 10 mins later (within the same github action)
- If a new deployment is attempted (new github action), cancel old github actions related to deployments.

References:
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28
- https://docs.github.com/en/actions/security-guides/automatic-token-authentication
- https://tldp.org/LDP/abs/html/recursionsct.html


btylerburton (approved):

nickumia-reisys:
Test successful 🎉 

image

PR #4454, opened 2023-09-06
`ids` should be interpreted as an array

nickumia-reisys:
Related to:
- https://github.com/GSA/catalog.data.gov/actions/runs/6098589650

References:
- https://unix.stackexchange.com/questions/615697/bash-iterate-array-detect-if-empty


FuhuXia (approved):

btylerburton (approved):
🧁

Issue #4488, opened 2023-10-11
Upgrade all NodeJS versions


btylerburton:
## User Story

In order to stay current with security & support, datagovteam wants to upgrade the node version on our static sites to a version supported by cloud.gov.

Ideally, we'd like to move to V20, but V18 will be supported through April 2025.
## Acceptance Criteria

[ACs should be clearly demoable/verifiable whenever possible. Try specifying them using [BDD](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications).]

- [ ] GIVEN [a contextual precondition] \
  [AND optionally another precondition] \
  WHEN [a triggering event] happens \
  THEN [a verifiable outcome] \
  [AND optionally another verifiable outcome]

## Background

[Any helpful contextual notes or links to artifacts/evidence, if needed]

## Security Considerations ([required](https://nvd.nist.gov/800-53/Rev4/control/CM-4))

High value.
Node versions should be kept current to take advantage of security patches.

## Sketch

- [x] Upgrade static sites (using Node V20 if possible, v18 if not)
  - [x] https://github.com/gsa/datagov-11ty
  - [x] https://github.com/gsa/resources.data.gov
  - [x] https://github.com/gsa/data-strategy
  - [x] https://github.com/gsa/us-data-federation

rshewitt:
the latest action for datagov-11ty succeeded despite Pa11y failing on the first URL ( privacy-policy ). need to update the [static site build template](https://github.com/GSA/data.gov/blob/main/.github/workflows/static-site-build-template.yml) used in the action file?

btylerburton:
related: https://github.com/GSA/data.gov/issues/4273

rshewitt:
acknowledging the issue with pa11y silently quitting during testing ( considering it a non-priority ) i'm going to proceed forward with upgrading the node versions without addressing this issue.

rshewitt:
resources.data.gov uses ruby 2.6 which isn't supported anymore. i'm gonna bump the ruby version.

rshewitt:
cloud.gov ruby upgrade [doc](https://cloud.gov/2023/05/24/cloud-gov-pages-jekyll-ruby-upgrade/) mentions 3.1. i'm gonna use 3.1.4.

rshewitt:
I'm checking the sites off in the sketch when the PR has been approved and merged.

PR #4513, opened 2023-10-25
update ruby version to equal resources. bump node version.

rshewitt:
# Pull Request

Related to [#4488](https://github.com/GSA/data.gov/issues/4488)

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/.github/workflows/static-site-qa-template.yml
+++ b/.github/workflows/static-site-qa-template.yml
@@ -12,7 +12,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.1.4'
+          ruby-version: '3.1'
           bundler-cache: true #runs `bundle install`
       - name: Install node.js
         uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/static-site-qa-template.yml
```diff
@@ -12,12 +12,12 @@ jobs:
         uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: '3.1'
+          ruby-version: '3.1.4'
```
jbrown-xentity: Are we sure we want to hardcode this to a patch version? Seems pretty specific, and will require a lot of watching and manual changes when releases happen...

jbrown-xentity (dismissed):

.github/workflows/static-site-qa-template.yml
rshewitt: I chose 3.1.4 because it's the last patch in 3.1 and I figured it's the most mature release of 3.1? that's a fair point i'll change it back.

rshewitt: changed in f4ee606f8d131f88c78e4be54787af0983a238e8

robert-bryson (approved):

PR #4579, opened 2024-01-05
if the app has 0 instances, don't restart it

FuhuXia:
GH Task `restart egress proxy` has been failing for the past a couple of weeks, due to timeout when restarting the erroneous 0/0 `logstack-space-drain` app. While the reason for the timeout is unknown, there is really no need to restart an app that has 0 instance running.

### DIFF
--- a/bin/restart-egress
+++ b/bin/restart-egress
@@ -2,9 +2,12 @@
 set -e
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


FuhuXia:
verified by a manual run 
https://github.com/GSA/data.gov/actions/runs/7424246666/job/20203451310

btylerburton (approved):

PR #5423, opened 2025-09-02
retry with new runner

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/pull/5417
- https://github.com/GSA/data.gov/pull/5421

## About

run the retry in a new runner, so that it gets run in a new ip, in case the IP address is blocked by AWS. 

## Context
https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1755790942340729

### DIFF
--- a/.github/workflows/app-restart-template.yml
+++ b/.github/workflows/app-restart-template.yml
@@ -21,91 +21,57 @@ on: # yamllint disable-line rule:truthy
         required: true
 
 jobs:
-  restart:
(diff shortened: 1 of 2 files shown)


FuhuXia:
It shows restart with the same runner does not work. The block is on ip level. need to use different runner to get a new ip.

image

cmhedrickREI (approved):
lgtm

PR #5439, opened 2025-09-05
retry restart in a new runner

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5415#issuecomment-3234562794

## About

figured out the reasons why previous PR did not work. This PR:
- retry restart in a new runner, hopefully it means the API is calling from a new ip therefore bypassing the AWS IP block.
- passing the environment secretes correctly. 
- uses `success.txt` in the artifact to mark result of restart attempt per app, so we only restart apps that fail for the first attempt, not the whole list of apps in the matrix.
- tested, it worked for restarting app if there is no 403. Need to deploy then capture a scenario with 403 error to verify the second restart attempt works. 

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/.github/workflows/app-restart-template.yml
+++ b/.github/workflows/app-restart-template.yml
@@ -14,42 +14,41 @@ on: # yamllint disable-line rule:truthy
         required: true
         type: string
 
-    secrets:
-      CF_SERVICE_USER:
-        required: true
-      CF_SERVICE_AUTH:
-        required: true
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
Good night is this complicated for just wanting to restart a machine. But let's try it and see if this solves our issue...

PR #5881, opened 2026-04-20
Add poetry npm

FuhuXia:
# Pull Request

handle npm build and poetry export


neilmb (approved):
LGTM (Do we still need these deploy fanciness as our app architecture gets simpler?)

PR #6070, opened 2026-06-17
Remove *.tigbox.com

FuhuXia:
This domain is added to prod trouble-shooting purpose. We are done with it.


jbrown-xentity (approved):
LGTM

PR #6103, opened 2026-06-24
exclude outbound link from pages not on catalog.data.gov

rshewitt:
# Pull Request

## About

11ty tests are [failing](https://github.com/GSA/datagov-11ty/actions/runs/28125174379/job/83287026460?pr=563) because there's blank outbound links from catalog-old which are registered by google analytics. the tests fail because the anchors are empty (the empty "linkUrl"). this filters out outbound links based on the referring origin being catalog.data.gov. this is one reason why we should consider turning off catalog-old. 

once this is merged i'll rerun metrics action then 11ty tests

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/metrics/datagov_metrics/ga.py
+++ b/metrics/datagov_metrics/ga.py
@@ -120,6 +120,16 @@ def setup_organization_reports():
                                 "stringFilter": {"matchType": "EXACT", "value": "true"},
                             }
                         },
+                        {
+                            "filter": {
+                                "fieldName": "pageLocation",
+                                "stringFilter": {
+                                    "matchType": "CONTAINS",
+                                    "value": "catalog.data.gov",
+                                    "caseSensitive": False,
+                                },
+                            }
+                        },
                         org_dimension_filter,
                     ],
                 },


FuhuXia:
why the snake_case not camelCase like other filters?

image

rshewitt:
> why the snake_case not camelCase like other filters?
> 
> image

I don't write GA4 expressions frequently and that was the example i saw. i tested locally and it filtered things as intended so it seems like both are supported. I'll update it to standardize.

rshewitt:
i know we don't have tests but here's my assertion. i'd prefer to not spend time setting up tests for this.

Screenshot 2026-06-25 at 9 39 11 AM

here's the fields so you can see that index 4 is the correct thing
```python
[
    {"name": "linkUrl"},
    {"name": "customEvent:DATAGOV_dataset_organization"},
    {"name": "customEvent:DATAGOV_dataset_publisher"},
    {"name": "outbound"},
    {"name": "pageLocation"},
    {"name": "pageTitle"},
]
```

FuhuXia (approved):

PR #6107, opened 2026-06-29
Gh actions

FuhuXia:
update actions/setup-node from v4 to v6 to remove the warning in the other repos when they use the snyk template

image


akuny (approved):
This looks good.

PR #6121, opened 2026-06-30
remove docker item from onboarding

FuhuXia:
No need for this for new members any more.
https://gsa-tts.slack.com/archives/C2N85536E/p1782849715037859


jbrown-xentity (approved):
LGTM

PR #6172, opened 2026-07-16
allow caller to specify ruby version

FuhuXia:
Allow callers to use different ruby versions, but keep 3.1 as default so it does not surprise old callers.

we need it when the caller app need to update ruby.


jbrown-xentity (approved):
LGTM

PR #6189, opened 2026-07-22
Fix: Update Broken Badge Link

akuny:
Updates an incorrect link in README.md: was `https://github.com/GSA/inventory-app/actions/workflows/deploy.yml`, should be `https://github.com/GSA/inventory-app/actions/workflows/publish.yml`.


jbrown-xentity (approved):
Pretty confusing that the publish.yml action is called "deploy"... https://github.com/GSA/inventory-app/actions/workflows/publish.yml

Good catch!

Issue #6205, opened 2026-07-28
O+M 07-28-26


cody-seibert-gsa:
As part of day-to-day operation of Data.gov, there are many [Operation and Maintenance (O&M) responsibilities](https://github.com/gsa/data.gov/wiki/Operation-and-Maintenance-Responsibilities). Instead of having the entire team watching notifications and risking some notifications slipping through the cracks, we have created an [O&M Triage role](https://github.com/gsa/data.gov/wiki/Operation-and-Maintenance-Responsibilities#om-triage-rotation). One person on the team is assigned the Triage role which rotates each sprint. _This is not meant to be a 24/7 responsibility, only East Coast business hours. If you are unavailable, please note when you will be unavailable in Slack and ask for someone to take on the role for that time._

Check the [O&M Rotation Schedule](https://docs.google.com/spreadsheets/d/1cF73HkzFuQth3z9voLfVO1Cc4kjBd6ecHGE2gkj5LCc/edit?usp=sharing) for future planning.

## Acceptance criteria
You are responsible for all [O&M responsibilities](https://github.com/gsa/data.gov/wiki/Operation-and-Maintenance-Responsibilities) this week. We've highlighted a few so they're not forgotten. You can copy each checklist into your daily report.

## Daily Checklist
- [ ] Review and address alerts sent to the `#datagov-alerts` Slack channel as they arrive.
- [ ] [Check Production State/Actions](https://github.com/GSA/data.gov#datagov-actions)
> **Note: Catalog Auto Tasks**
> You will need to update the chart values manually. Click the Action link in each issue and grab the values from `monitor task output` and `check runtime`.
- [ ] [Check auto generated O&M tickets](https://github.com/orgs/GSA/projects/11?pane=issue&itemId=47716346) from no status column
- [ ] Check [Harvesting Emails](https://github.com/gsa/data.gov/wiki/Operation-and-Maintenance-Responsibilities#harvest-job-report-daily-email-report)
- [ ] [New Relic Alerts](https://alerts.newrelic.com/accounts/1601367/incidents) Triaged
- [ ] Check each github repo's Security tab to traige issues found by Dependabot and Code scan

## Weekly Checklist
[...]

PR #6206, opened 2026-07-28
Update README.md

cody-seibert-gsa:
work for #6205

removing ssb from the readme as it's not longer used


FuhuXia:
there are two two ssb rows.

jaredb96 (approved):
LGTM

PR #6219, opened 2026-08-04
Remove Check-Db-Models badge link in README.md

marcos-nieto-usds:
This action was deleted when we moved our data access functionality. Therefore, the badge https://github.com/GSA/datagov-catalog/actions/workflows/check-db-models.yml/badge.svg was broken and is no longer needed because we’re not going back to that approach anymore of comparing shared db code between catalog and harvester.


jbrown-xentity (approved):
LGTM

